PR #1, opened 2016-05-03
Add blacklist for irc users (untested)

thenaterhood:
This adds a blacklist that allows for ignoring a list of users from the IRC side. I have not tested this (I'm in a meeting) but it's super simple so it should be quick to check and fix if it's a problem.


repkam09:
Confirmed working! Thanks!

PR #4, opened 2016-09-06
Move all options into the config file

repkam09:
I figured this might be a commit worth doing. Move all of the config options into the config.json file instead of having to modify the source code file just for options.

Added some new flags for feature toggles so that you can turn on and off leave/join messages and to customize the spacer in telegram messages when they are mirrored into IRC. Someone was saying they didnt like the ":" and thought it should be ">" or something else.

I didnt actually test any of this... but thought I would dump the code up here to see if it seemed like a reasonable idea.


Serubin:
What if we had suffix & prefix options?
< nick > message

jflory7:
I'm definitely a big +1 for separating this information out into the config file. I think this will also make it easier to apply updates to the bot over time as well. I haven't tested it yet, but I'm willing to test it out in prod later…

jflory7:
Oh, also, @repkam09, you might need to update the README docs to correspond with these changes as well.

config.json.example
```diff
@@ -1,3 +1,19 @@
...
+    },
+    "tg": {
+        "chatId": "-0000000000000",
+        "spacer": ":",
+        "showJoinMessage": true,
+        "showActionMessage": true,
+        "showLeaveMessage": false,
+        "showKickMessage": false
+    }
 }
```
robbyoconnor: you _COULD_  rename this to .js -- and simply export the JSON object -- by simply requiring it, you get a javascript object.

teleirc.js
```diff
@@ -2,25 +2,95 @@ var tg = require("node-telegram-bot-api");
 var irc = require("irc");
 var fs = require("fs");
```
robbyoconnor: use `const` from ES6

teleirc.js
```diff
@@ -2,25 +2,95 @@ var tg = require("node-telegram-bot-api");
...
 
+// If you want to change the default config file name, do that here.
+var configfilename = 'config.json';
```
robbyoconnor: use `const` from ES6

teleirc.js
```diff
@@ -2,25 +2,95 @@ var tg = require("node-telegram-bot-api");
...
+
+// This is the base config object that will store values from the settings file.
+var config = {};
```
robbyoconnor: use `let` from ES6

teleirc.js
```diff
@@ -2,25 +2,95 @@ var tg = require("node-telegram-bot-api");
...
+
 // Read in the config file to hide the bot API token
 console.log("Reading config.json file to get bot API token...");
-var settings = JSON.parse(fs.readFileSync('config.json', 'utf8'));
-if (settings.token) {
-    console.error("Token value not found in config.json file");
+var settings = JSON.parse(fs.readFileSync(configfilename, 'utf8'));
```
robbyoconnor: use `const` from ES6

teleirc.js
```diff
@@ -2,25 +2,95 @@ var tg = require("node-telegram-bot-api");
...
+         config.showActionMessage = settings.tg.showActionMessage;
+    } else {
+        config.showActionMessage = false;
+    }
+
+} else {
+    setupError("Unable to find Telegram settings in " + configfilename);
+}
+
+var token = null;
```
robbyoconnor: use `let` from ES6

teleirc.js
```diff
@@ -73,8 +143,6 @@ tgbot.on('message', function (msg) {
 
 // Action to invoke on incoming messages from the IRC side
 ircbot.addListener('message', function (from, channel, message) {
-    // Anything coming from IRC is going to be valid to display as text
-    // in Telegram. Just do a quick passthrough. No checking.
     var matchedNames = config.ircBlacklist.filter(function (name) {
```
robbyoconnor: use `let` from ES6

teleirc.js
```diff
@@ -85,37 +153,51 @@ ircbot.addListener('message', function (from, channel, message) {
...
 
 
-// Let the telegram chat know when a user joins the IRC channel
-ircbot.addListener('join', function (channel, username) {
-    tgbot.sendMessage(config.chatId, username + " has joined " + config.channel + " channel.");
-});
+// These additional alerts can be turned on in the config file
+if (config.showActionMessage) {
+    ircbot.addListener('action', function (from, channel, message) {
+        var matchedNames = config.ircBlacklist.filter(function (name) {
```
robbyoconnor: use `let` from ES6

robbyoconnor: Use arrow functions: 
```js
   ircbot.addListener('action', (from, channel, message) => { 
      ... 
   }); 
```

teleirc.js
```diff
@@ -85,37 +153,51 @@ ircbot.addListener('message', function (from, channel, message) {
...
-*/
+        if (matchedNames.length <= 0) {
+            tgbot.sendMessage(config.chatId, from + " " + message);
+        }
+    });
+}
+
+if (config.showJoinMessage) {
+    // Let the telegram chat know when a user joins the IRC channel
+    ircbot.addListener('join', function (channel, username) {
```
robbyoconnor: Same here 
```js
 ircbot.addListener('join', (channel, username) => {
     ... 
 }); 
```

robbyoconnor:
So I made comments but got lazy -- convert all callbacks to use arrow functions -- assuming you're running node 4 or better... I gave you examples in two parts.

robbyoconnor:
@repkam09, I edited some of my comments, have a look...your code just got much, much simplier.

ritjoe:
> Someone was saying they didnt like the ":" and thought it should be ">" or something else.
A common behavior in IRC clients is tab-autocompletion for nicks in the channel, used as a means to highlight (notify) their users through the highlighted user's client.

eg, for me to direct something to Serubin's attention I might issue:

```
Ser<tab>
```
which will be completed to 
```
Serubin: 
```

And then after the ":" I would type my message to his attention (but still publically in channel, as opposed to a personal message).

Seeing a username followed by a colon from the tg bridge makes it look (at least, to an IRC user) like it's a message **for** that user instead of, as is the current case, a message **from** that user.

Serubin:
I agree. That's why I propose the suffix/prefix.
The format `< teleirc > < nick > message` might be clearer.

repkam09:
Added most of the recommendations given by @robbyoconnor (thanks!). A bit more to come later with the new feature requests, that will have to be done after 5pm. :)

robbyoconnor:
One of the biggest perks of doing things in js files is you can store secrets in environment variables and use 

`process.env.FOO || "some default"` (if the first is null, then the second happens)

PR #8, opened 2016-09-21
Bold username on IRC -> Telegram message

tjhorner:


tjhorner:
Something that you might want to look into though: Markdown escaping. Telegram freaks the fuck out if you give it invalid Markdown, so someone could theoretically put some invalid markdown in their message and kill it.

tjhorner:
Alternatively, you could use HTML since it's easier to escape. (change `parse_mode` to "HTML")

repkam09:
Sounds like a reasonable addition to me. HTML is probably a better choice as people are less likely to use that in a chat message compared to something that might look like Markup.

tjhorner:
@thenaterhood I can add that when I get home. First, I'm going to test if Telegram even supports &escaped; HTML entities.

thenaterhood:
It looks like node has markdown escaping resources as well, so I'd say with an escaped message, markdown would be okay. A quick search turned up https://www.npmjs.com/package/markdown-escape

tjhorner:
@thenaterhood The reason I'm saying HTML would be easier is that Telegram's version of markdown is really different and the package you mentioned would escape where not needed (like in strikethroughs)

thenaterhood:
@tjhorner something we might need to decide is if we want to have telegram render markdown from IRC messages at all. It would be a different experience on the Telegram side as IRC would see only plaintext, and I could see it adding some confusion depending how it renders, since we may start bundling multiple messages together to work with Telegram's rate limiting.

tjhorner:
@thenaterhood I don't want markdown from users to render, it would look something like this:

### From IRC

> tjhorner: hey guys
> tjhorner: seen the new update?
> testuser: yeah, it's pretty neat
> tjhorner: it adds \*\*markdown\*\* too, pretty cool

### From Telegram

> Bot:
> **tjhorner**: hey guys
> **tjhorner**: seen the new update?
> **tjhorner**: yeah, it's pretty neat
> **tjhorner**: it adds \*\*markdown\*\* too, pretty cool

(e.g. the markdown won't be rendered inside of the user's message body, just their username.)

And also, I did some digging and it seems the Telegram API supports HTML escaping with a couple of HTML entities:

> - All numerical HTML entities are supported.
> - The API currently supports only the following named HTML entities: `&lt;`, `&gt;`, `&amp;` and `&quot;`.

So we could just do a simple find-replace regex on the message body to prevent users from putting in their own HTML into the message on Telegram.

tjhorner:
I guess I should actually clone my fork of the repo so I don't need to use the web editor, lol

thenaterhood:
Yeah I get what you're going for and I like the idea. Don't write your own
escaping though, use an existing library for it. What already exists will
be safer and often more efficient.

tjhorner:
The thing is, if it escapes a character we don't want escaped, Telegram will throw errors. Are there libraries that allow you to only escape certain characters?

thenaterhood:
Escaping more than needed wouldn't cause any issues in this case. If we're sending markdown, we need to escape all markdown we don't want rendered. The string concatonation would be something like 

`*username*: + escape(message)`

tjhorner:
Oh, we're on different pages here. I was talking about HTML, but yeah what you're saying makes sense. Maybe we should stop discussing this and I should actually push the code haha

thenaterhood:
@tjhorner Any updates on this?

robbyoconnor:
Ignore my approval.

tjhorner:
@thenaterhood no, sorry. School has been a burden and stuff, but I can probably do it today.

tjhorner:
@robbyoconnor sounds good.

jflory7:
@tjhorner You may also need to rebase to pull in some of the newest changes from the master branch.

jflory7:
If we can get this by Sunday afternoon, we can try to cut this into a 1.1.0 release. :smile:

jflory7:
# Closing pull request

Since we've been merging in other changes and this pull request has become a little messy, I'm going to make the decision to close it out. @tjhorner, should you have time to revisit this, feel free to either reopen this PR or file a new one. I would like to see this functionality added in.

I'm going to rule this feature out for the 1.1.0 release, but if you resubmit, we can definitely get it added to 1.1.1. :smile: Hope you will have time to work on this again!

Issue #12, opened 2016-10-20
Bot fails to post to Telegram


bmamlin:
I was creating a Docker image from this repo and it failed to post to Telegram. Debugging `teleirc.js` revealed a couple places where `chatId` was mistyped as `chatID`. A pull request is on its way!

thenaterhood:
Resolved by #13.

PR #13, opened 2016-10-20
Fix case of chatId references

bmamlin:
Here's a quick fix for #12


thenaterhood:
Ayyy, good catch! Looks all good to me!

PR #14, opened 2016-10-21
Add support for Docker

bmamlin:
This adds support for running teleirc via Docker or natively. If you are interested in this, you will want to set up your own repo on [Docker Hub](https://hub.docker.com/) so people can always get the latest version of teleirc running with a single command:
1. If you don't already have an account, sign up for a free account at https://hub.docker.com/ (e.g., RITlug)
2. Link your Docker Hub account to your GitHub account ([instructions](https://docs.docker.com/docker-hub/builds/))
3. Change the image reference in the `docker run` example in README.md from "burke/teleirc" to match your Docker Hub account (e.g., "RITlug/teleirc")
4. Merge this pull request
5. Log into Docker Hub and select Create > Create Automated Build in the top right to define a public "teleirc" image
6. Under [teleirc github settings](https://github.com/RITlug/teleirc/settings), go to Integrations & services and add "Docker" service to trigger automatic builds of the Docker Hub image when commits are made to github.

Once that's done, anyone can run teleirc within a single docker run command by passing in their configuration via environment variables. 😃 

On the other hand, if you don't want to support Docker in this way, just close this PR and I'll understand.

PR #22, opened 2016-10-28
Add support for Docker

itamarjp:
patches from bmamlin


robbyoconnor:
Switch to alpine as specified here: https://github.com/RITlug/teleirc/pull/14#issuecomment-257024886

jflory7:
As discussed in #14, we will merge this in since it has the requested changes from that PR.

PR #25, opened 2016-11-07
This is a simple change which simply changes var to let in config files

robbyoconnor:
It's not a HUGE thing but it bugs me.


robbyoconnor:
This change is now deprecated in #26 -- since I completely nuked config.js.docker -- it's not DRY.

PR #26, opened 2016-11-07
[RDY] Docker improvements

robbyoconnor:
This PR does various things all centered around Docker improvements -- a lot of security but an addition too!
Changes: 
- Make the teleirc not run as root in the Docker images
- add a dockerignore to not bake in everything.
- Add a table comparing the Docker image sizes for educated decisions
- Nuke config.js.docker in favor of config.js.example
- Add a docker-compose.yml example file and document it in the README
- Add dotenv for reuse of the .env file from docker-compose for running
  natively
- Improvements to the README around configuration
Closes #24

PR #27, opened 2016-11-08
Add support for the 'edited_message' telegram event

repkam09:
Small change to add support for the 'edited_message' telegram event so if someone on the Telegram side edits a message, that edit is then sent to the IRC side. Before, the edit would happen on the Telegram side (as usual) but the change would not be sent to the IRC side.

To support this change, the node telegram bot npm package was bumped to the latest version which supports this new event.


lib/libteleirc.js
```diff
@@ -208,11 +208,12 @@ class TeleIrc {
         this.tgRateLimiter = new MessageRateLimiter(
             this.config.tg.maxMessagesPerMinute,
             60,
-            function(message) {
+            function (message) {
```
robbyoconnor: Use arrow functions

robbyoconnor: It's much cleaner here.

robbyoconnor:
you are going to need to rebase after #26 is merged...one of us will have to..

jflory7:
@repkam09 Would you mind rebasing so we can merge this in? :) Also, any idea why your PR went straight to the `master` branch instead of `devel`?

robbyoconnor:
I caused the conflict by adding `dotenv` -- sorry :)

thenaterhood:
Does this also add support for passing files to IRC? The end of the diff looks like it does.

jflory7:
Whoops. Any reason why we can't merge this into the repository? Seems like we're all good?

robbyoconnor:
After it's squashed because I'm OCD about revision history @jflory7

robbyoconnor:
I think you can squash it when you merge it actually...and probably should.

thenaterhood:
GitHub pulls can do squash and merge. That's generally preferred when
merging into a dev branch (but not to master/prod; causes headaches).

robbyoconnor:
You may wanna rebase to `devel`?

jflory7:
@robbyoconnor Yeah, I noticed that it went to `master` and I was working on rebasing locally, but have some merge conflicts I needed help on solving. I shared more info in our Telegram group.

PR #23, opened 2016-10-31
First Unit Tests Online

xforever1313:
This pull request puts the first unit tests online.  It also fixes a few potential bugs in the configuration validation:

- Fixed an issue where if sendStickerEmoji, showJoinMessage. showLeaveMessage were NOT in the settings, were undefined, or null, we would crash.  We now set these to default values instead of crashing.

- Added an TeleIrcException class, along with an Error code class.

- Added default settings to the TeleIrc Class.

- Added unit tests for the IrcConfig validation.  Unit tests ensure that if required settings are missing, are null, or are undefined, we throw Exceptions.  Anything that is option is set to the default if they are null, undefined, or missing.

To run unit tests, install nodeunit, and on the command line run in the tests directory "nodeunit ./".

PR #28, opened 2016-11-11
Config test cherrypick

xforever1313:
Repull of #23, since I suck at git and can't handle conflicts :|.

Same changes, different pull.


robbyoconnor:
why not squash this?

thenaterhood:
I think we have squashes disabled on the repo for some reason, but a squash and merge would be the best approach for this. There's still a conflict here for some reason though that I'm looking into.

thenaterhood:
Github has lied to me, apparently. There is no longer a conflict now that I've refreshed the page. Wut. All good by me though and it's the same changes that were waiting for my approval in #23 so I'm pulling the trigger.

robbyoconnor:
also, I can teach you how to resolve conflicts -- it's not that bad :)

PR #29, opened 2016-11-11
Add travis config (continuous integration)

thenaterhood:
Let's get that continuous integration started! It appears Travis is already trying to build changes.

We can also configure this so that pull requests are automatically tested and Travis posts the results on the pull request itself.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,6 @@
+language: node_js
+install:
+        - npm install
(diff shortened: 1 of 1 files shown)


thenaterhood:
This is a new file that doesn't impact the application. You can also see proof that it's okay with the checks on this pull. Merging!

PR #38, opened 2017-10-02
Uploaded Photos appear in IRC

xforever1313:
This commit will post the URL of photos uploaded to the Telegram Group to IRC.

Note, however, the URL that is posted will only last an hour before the link expires.  A more long-term solution will take more time to implement.  This first step is to get the URL so we can use it for those other long-term solutions.

### DIFF
--- a/lib/libteleirc.js
+++ b/lib/libteleirc.js
@@ -77,7 +77,6 @@ class TeleIrc {
         this.defaultIrcSuffix = "";
         this.defaultShowJoinMessage = false;
         this.defaultShowLeaveMessage = false;
-
     }
 
     // ---------------- Functions ----------------
@@ -244,7 +243,10 @@ class TeleIrc {
             });
 
             if (matchedNames.length <= 0) {
-                this.queueTelegramMessage(this.config.tg.chatId, "<" + from + "> " + message);
+                this.queueTelegramMessage(
+                    this.config.tg.chatId,
+                    "<" + from + "> " + message
+                );
             }
         });
 
@@ -260,7 +262,10 @@ class TeleIrc {
                 });
 
                 if (matchedNames.length <= 0) {
-                    this.queueTelegramMessage(this.config.tg.chatId, from + " " + message);
+                    this.queueTelegramMessage(
+                        this.config.tg.chatId, 
+                        from + " " + message
+                    );
                 }
             });
         }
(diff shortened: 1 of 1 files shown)


lib/libteleirc.js
```diff
@@ -367,22 +396,34 @@ class TeleIrc {
...
-                        this.ircbot.say(this.config.irc.channel, this.config.irc.prefix + from + this.config.irc.suffix + " " + emoji);
+                        this.sendUserMessageToIrc(from, emoji);
                     }
                 } else {
                     // Check if this message contains a document/file to display in irc
                     if (msg.document) {
                         this.tgbot.getFileLink(msg.document.file_id).then((url) => {
-                            let newmessage = sg.document.file_name + " (" + msg.document.mime_type + ", " + filemsg.document.file_size_size + " bytes) : " + url;
-                            this.ircbot.say(this.config.irc.channel, this.config.irc.prefix + from + this.config.irc.suffix + " " + newmessage);
+                            let newmessage = msg.document.file_name + " (" + msg.document.mime_type + ", " + msg.document.file_size_size + " bytes) : " + url;
```
jflory7: Why change `sg` => `msg` and `filemsg` => `msg`?

lib/libteleirc.js
```diff
@@ -416,6 +457,49 @@ class TeleIrc {
...
+            }
+            else if (photo.file_size < innerPhoto.file_size) {
+                photo = innerPhoto;
+            }
+        }
+
+        return photo;
+    }
+
+    getPhotoMessage(caption, from, fileUrl) {
```
jflory7: 🎉 💯 🎉

lib/libteleirc.js
```diff
@@ -433,6 +517,19 @@ class TeleIrc {
...
         }
     }
+
+    /**
+     * Sends the given Telegram message from a USER to the IRC channel.
+     * This includes the IRC prefix and the suffix.
+     * @param {String} from - Who the message was from on the Telegram Side. 
+     * @param {String} message - The message to send to the IRC channel.
+     */
+    sendUserMessageToIrc(from, message) {
```
jflory7: I was wondering where this method came from, didn't see it until I worked my way to the bottom. Nice idea, this is a little more clean. 👍

lib/libteleirc.js
xforever1313: sg I think was a typo... it should have been msg all along.  If we tried uploading a document, teleirc crashed.  I have no idea where filemsg came from.. my guess is that was also a typo.

Issue #33, opened 2016-12-13
One should be able to see files / images shared from telegram


KevinMGranger:

jflory7:
I believe this is possible. I remember @repkam09 mentioned the possibility of doing it, but the files "expire" after some time, unless we were to upload them to another site or service, like Imgur for images.

repkam09:
I wrote up a quick version of this which posts the link directly to the Telegram (temporary, 1 hour) version of the file. Having it upload somewhere more permanent would be good, but I'm not sure where that would be.

If you have a simple file repository on the RITLug servers that can be hit from our service here, maybe. I have done upload scripts using node and restify for my own personal projects. But then comes the issue of protecting that from random people uploading file, setting limits, deleting after some amount of time, etc.

robbyoconnor:
TeleIRC (the other one) spins up a web server and hosts it on the server...

mestaritonttu:
Why not join forces with the other teleirc? Is there some fundamental technological disagreement preventing such a merger?

jflory7:
@mestaritonttu It was mostly that we created this project for our own community, then it ended up becoming a little bit popular, and we kept maintaining it. There's no disagreement, but more that we're just maintaining what we created and probably don't have the time to actively contribute to many more projects.

xforever1313:
Created this pull https://github.com/RITlug/teleirc/pull/38 to get Telegram's URL to the uploaded photo.  Right now, it just posts the URL to the IRC channel, but as Mark said, that URL only lasts an hour.  But, we need to get the URL before we can do a more persistent solution.

jflory7:
@xforever1313 Amazing! This is still a lot of progress. We could potentially try to link to an API for a site like Imgur to upload the photos and post that link instead. Will try to review the PR soon!

Chris-Bitler:
The Imgur API for uploading images supports passing it a URL, so perhaps it could be passed the telegram URL? https://apidocs.imgur.com/#4b8da0b3-3e73-13f0-d60b-2ff715e8394f

Chris-Bitler:
Just an update on this: I am working on implementing uploading images to imgur, but they have some bug with their API application registration that caused it to auto redirect away from the page; I was told by their support that this will be fixed on monday.

Chris-Bitler:
I'm not sure how commenting did that.

jflory7:
[...]

PR #39, opened 2017-10-09
Add ability to upload images from telegram API to imgur and have the bot use the imgur link

Chris-Bitler:
This is related to the discussion in https://github.com/RITlug/teleirc/issues/33 - Since telegram doesn't send the correct mime type for the file uploaded (it sends application/octet-stream instead), in order to allow people to view the images uploaded without having to download them to their own device, it makes sense to re-upload them somewhere else.

In order for this to work, people who intend to use the imgur upload functionality need to use imgur's API site to request a clientID and replace it in the .env file. This is why imgur upload is disabled by default.


Chris-Bitler:
I will fix the tests that this broke

jflory7:
@Chris-Bitler See if you can fix the build errors and we'll get this merged in. Also, thanks for writing a brief snippet of docs to add into the README. This is helpful.

Also, while the environment variables are nice if you're using Docker, we should also have a way for people to use the configuration file, `config.js`, to configure the Imgur support. Can you make sure that this functionality is included into `config.js.example`?

.env.example
```diff
@@ -12,4 +12,6 @@ SHOW_ACTION_MESSAGE=true
 SHOW_JOIN_MESSAGE=true
 SHOW_KICK_MESSAGE=true
 SHOW_LEAVE_MESSAGE=true
+USE_IMGUR_FOR_IMAGES=false
+IMGUR_CLIENT_ID=0000000000
```
jflory7: Also, while the environment variables are nice if you're using Docker, we should also have a way for people to use the configuration file, `config.js`, to configure the Imgur support. Can you make sure that this functionality is included into `config.js.example`?

Chris-Bitler:
Tests pass now. I didn't add any tests related to exceptions for the imgur section, because it fails gracefully if it doesn't exist and just uses the telegram file url.

jflory7:
I rebased these commits into ecf15a1b1892373f57314eccd4068d706083f7e5 and pushed it to devel. This is now live! Thanks for contributing, @Chris-Bitler! 🎉 💯 🎉

🌮 🌯

PR #40, opened 2017-10-10
[RDY] A lot of refactoring and bug fixes

robbyoconnor:


robbyoconnor:
@Chris-Bitler -- Done.

config.js.example
```diff
@@ -6,23 +6,23 @@ let settings = {
         server: process.env.IRC_SERVER || "irc.freenode.net",
         channel: process.env.IRC_CHANNEL || "",
         botName: process.env.IRC_BOT_NAME || "teleirc",
-        sendStickerEmoji: process.env.IRC_SEND_STICKER_EMOJI || true,
+        sendStickerEmoji: process.env.IRC_SEND_STICKER_EMOJI === 'true'|| true,
```
Chris-Bitler: Huge nitpick from a consistency standpoint, but this should probably be using either single quotes or double quotes and keeping it consistent (it uses single quotes for this and the two below it then switches to double quotes)

robbyoconnor: I missed that one!

robbyoconnor:
@jflory7 I ran [js-beautify](https://github.com/beautify-web/js-beautify) across the source files. 

I also tweaked the Docker User/Group Ids to be lower.

Dockerfile.slim
```diff
@@ -1,6 +1,6 @@
 FROM node:6-slim
 
-RUN groupadd teleirc -g 99999 ; useradd -u 99999 -g teleirc -s /bin/bash -M -d /opt/teleirc teleirc
+RUN groupadd teleirc -g 65533 ; useradd -u 65533 -g teleirc -s /bin/bash -M -d /opt/teleirc teleirc
```
jflory7: @robbyoconnor As a reference, could you explain why you changed the group and user IDs? I saw the commit e7396255ede70e83f8334cb5f0c8f3cc39a438e9 but wasn't sure why this needed to be done.

config.js.example
```diff
@@ -1,29 +1,31 @@
...
+    prefix: process.env.IRC_PREFIX || "<",
+    suffix: process.env.IRC_SUFFIX || ">",
+    showJoinMessage: process.env.IRC_SHOW_JOIN_MESSAGE === "true" || true,
+    showLeaveMessage: process.env.IRC_SHOW_LEAVE_MESSAGE === "true" || true,
+  },
+  tg: {
+    chatId: process.env.TELEGRAM_CHAT_ID || "-000000000",
+    showJoinMessage: process.env.SHOW_JOIN_MESSAGE === "true" || false,
+    showActionMessage: process.env.SHOW_ACTION_MESSAGE === "true" || true,
+    showLeaveMessage: process.env.SHOW_LEAVE_MESSAGE === "true" || false,
```
jflory7: @robbyoconnor For clarity, the environment variables in `env.example` sets both the Telegram and IRC settings? It would be nice if we could separate them.

Dockerfile.slim
Serubin: @jflory7 65533 is the largest group ID that is available in the linux kernel.

Serubin:
@jflory7 In regards to `.env.example` file. I believe the default location is `.env` thus `.env.example` would be appropriate. `env.example` would lead people to believe that the file should be `env` rather than `.env`.

config.js.example
robbyoconnor: Yup. process.env is populated by dotenv very early in execution.

robbyoconnor:
@Serubin yup with an added test to ensure that the config returns the proper types and not just strings.

Dockerfile.slim
robbyoconnor: Yeh that was my screw up -- and I own it.

jflory7:
As a note, I've asked @repkam09, @xforever1313, @Chris-Bitler, and @Serubin to do a code review on this PR. Since it's originally @repkam09's project, I'd like to get his approval + one more from anyone before merging (as well as addressing the change requests above).

.jsbeautifyrc
```diff
@@ -0,0 +1,18 @@
+{
+  "indent_size": 2,
```
Chris-Bitler: I understand that this is following the airbnb style guide, but this is just my personal opinion: Only having two spaces makes the code harder to read, especially with the callback hell that javascript can be.

.travis.yml
```diff
@@ -5,10 +5,13 @@ cache: yarn
...
-
+  - 7
+  - 8
+before_install:
+  - cp .env.example .env
+  - cp config.js.example config.js
 install:
-  - npm install
-  - npm install -g nodeunit
+  - yarn
```
Chris-Bitler: This is more a question than anything else - why the switch from NPM to yarn?

robbyoconnor:
I squashed commits as much as I could -- some things really don't belong clumped into one commit.

robbyoconnor:
OH and we now test against node 7 and 8 as well @Serubin

Serubin:
@robbyoconnor Awesome!

Dockerfile.slim
jflory7: Okay, cool, thanks for the clarification. 😁 Makes sense then!

.travis.yml
xforever1313: Just out of curiosity, why choose yarn over npm?  I personally don't have a preference, just curious.

jflory7:
After the last few comments and squashed commits, everything looks good to me. 👍 Updating review to approved.

.travis.yml
jflory7: In retrospect, I'm also curious about this.

robbyoconnor: Caching.

robbyoconnor: Yarn keeps a cache on travis.

lib/TeleIrcException.js
```diff
@@ -0,0 +1,23 @@
...
+  /**
+   * Constructor.
+   * @param errorCode - Unique number for this error.
+   * @param message - Corresponding Message.
+   */
+  constructor(errorCode, message) {
+    this.errorCode = errorCode;
+    this.message = message;
+  }
+
```
xforever1313: This was probably missing in the first place before the move, but we should put

// ---------------- Functions ----------------

Here to go here so we are consistent with other files.

robbyoconnor: I can add that back -- I ran js-beautify on them

.jsbeautifyrc
xforever1313: I second this...  2 spaces seems too small imo...

robbyoconnor: You have no reason to even be dealing with that. async/await works in node 6,7 and 8. You can use promises...which eliminates callback hell. To me,  4 spaces just looks like wasted space.

lib/TeleIrcException.js
robbyoconnor: Fixed.

.travis.yml
repkam09: Anecdotal, but I have also been using yarn for a while now and found it much faster and predictable than npm. The yarn lock file helps, but npm has that now as well with v5. The speed is real though.

robbyoconnor:
@repkam09 -- There are tests that cover the actual original intent of this PR

lib/TeleIrcException.js
robbyoconnor: @xforever1313 also re: the functions thing -- No need... seems silly.

.travis.yml
robbyoconnor: Yeah builds on travis are WAY quicker...esp. due to the caching.

PR #42, opened 2017-12-01
Fix Dockerfile

mbologna:


jflory7:
Oh, nice catch, @mbologna. Thanks for submitting this PR! Merging. 🎬

PR #65, opened 2018-08-08
Nickserv identify

michalrud:
Bot will `IDENTIFY` itself to NickServ (or other compatible service) after connecting to IRC.

Bonus functionality:
- Messages sent from user with nick configured as `IRC_NICKSERV_SERVICE` will be ignored,
- Messages sent anywhere but on the configured channel are ignored
    - Rationale: someone can spam direct messages at bot and it will be forwarding all that to the telegram group,
    - But mainly: if IRC channel has `+r` and redirection to other channel in case of lack of identification and get forwarded as it wasn't yet identified, it would forward messages from that channel even if this is not what was intended. This could be done better, but there are two solutions I see:
        - A small timeout, after which we assume to be identified and join the intended channel. Kinda fishy,
        - Parsing messages from `NickServ`. Would assume, that NickServs on all servers would behave in the same way. Or we assume we're always on Freenode and hardcode it (or better - leave a regex that can be configured).

### DIFF
--- a/lib/TeleIrc.js
+++ b/lib/TeleIrc.js
@@ -190,7 +190,8 @@ class TeleIrc {
    * Identifies to NickServ, if password is provided.
    */
   nickservIdentify() {
-    if (this.config.irc.nickservPassword.length != 0) {
+    if ((this.config.irc.nickservPassword.length > 0) &&
+        (this.config.irc.nickservService.length > 0)) {
       console.log("Identifying to NickServ...");
       this.ircbot.say(
         this.config.irc.nickservService,
@@ -212,7 +213,7 @@ class TeleIrc {
         return from.toLowerCase() === name.toLowerCase();
       });
 
-      if ((this.config.irc.nickservService.length != 0) &&
+      if ((this.config.irc.nickservService.length > 0) &&
           (from.toLowerCase() === this.config.irc.nickservService.toLowerCase())) {
         console.log(`[${this.config.irc.nickservService}] ${message}`);
         return; // we don't want nickserv messages being forwarded to telegram
--- a/lib/TeleIrc.js
+++ b/lib/TeleIrc.js
@@ -214,8 +214,14 @@ class TeleIrc {
 
       if ((this.config.irc.nickservService.length != 0) &&
           (from.toLowerCase() === this.config.irc.nickservService.toLowerCase())) {
-        console.log("[NickServ] " + message);
+        console.log(`[${this.config.irc.nickservService}] ${message}`);
         return; // we don't want nickserv messages being forwarded to telegram
+      }
+
+      // we want to ignore the messages from not configured channels or direct messages
+      if (channel !== this.config.irc.channel) {
+        console.log(`Ignoring: <${from}@${channel}> ${message}`);
+        return;
       }
 
       if (matchedNames.length <= 0) {
--- a/lib/TeleIrc.js
+++ b/lib/TeleIrc.js
@@ -193,9 +193,13 @@ class TeleIrc {
     if ((this.config.irc.nickservPassword.length > 0) &&
         (this.config.irc.nickservService.length > 0)) {
       console.log("Identifying to NickServ...");
+      var nickservRegisteredNick = this.config.irc.botName;
+      if (this.config.irc.nickservRegisteredNick.length > 0) {
+        nickservRegisteredNick = this.config.irc.nickservRegisteredNick;
+      }
       this.ircbot.say(
         this.config.irc.nickservService,
-        "IDENTIFY " + this.config.irc.nickservPassword
+        "IDENTIFY " + nickservRegisteredNick + " " + this.config.irc.nickservPassword
       );
     } else {
       console.log("NickServ password not provided - omitting indentification");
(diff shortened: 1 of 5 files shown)


lib/TeleIrc.js
```diff
@@ -193,6 +212,18 @@ class TeleIrc {
         return from.toLowerCase() === name.toLowerCase();
       });
 
+      if ((this.config.irc.nickservService.length != 0) &&
```
thenaterhood: This is pretty much just pedantics, but typically for things like this I like to check for greator than 0, instead of != 0. If you look down on line 227, we're using that check (though we're using the inverse).

lib/TeleIrc.js
```diff
@@ -180,6 +180,25 @@ class TeleIrc {
...
+  }
+
+  /**
+   * Identifies to NickServ, if password is provided.
+   */
+  nickservIdentify() {
+    if (this.config.irc.nickservPassword.length != 0) {
+      console.log("Identifying to NickServ...");
+      this.ircbot.say(
+        this.config.irc.nickservService,
```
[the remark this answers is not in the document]
michalrud: That's right, I forgot about it. Thanks!

Tjzabel:
Hm, this should also fix #66

lib/TeleIrc.js
```diff
@@ -180,6 +180,26 @@ class TeleIrc {
...
+  /**
+   * Identifies to NickServ, if password is provided.
+   */
+  nickservIdentify() {
+    if ((this.config.irc.nickservPassword.length > 0) &&
+        (this.config.irc.nickservService.length > 0)) {
+      console.log("Identifying to NickServ...");
+      this.ircbot.say(
+        this.config.irc.nickservService,
+        "IDENTIFY " + this.config.irc.nickservPassword
```
jwflory: In some cases, the bot has issues "claiming" a nick (where `my-bot` becomes `my-bot1`, `my-bot2`, etc.), so the given nickname may be different than a registered nick. We have not figured out why this happens yet.

Could we add a fixed configuration parameter for a NickServ username, different from the bot name? The syntax to send to a NickServ account would be:

`IDENTIFY <NickServ username> <NickServ password>`

umlaeute:
it seems that there is some race, that does a `JOIN` *before* the `IDENTIFY` succeeds:

~~~
Setup complete! Teleirc now running.
31 Aug 19:00:43 - Sending irc NICK/USER
31 Aug 19:00:43 - SEND: NICK <USERNAME>
31 Aug 19:00:43 - SEND: USER nodebot 8 * :nodeJS IRC client
Identifying to NickServ...
31 Aug 19:00:43 - SEND: PRIVMSG NickServ :IDENTIFY <PASSWORD>
31 Aug 19:00:43 - SEND: JOIN #channel
31 Aug 19:00:43 - MODE: <USERNAME> sets mode: +Ri
31 Aug 19:00:43 - GOT NOTICE from "NickServ": "This nickname is registered. Please choose a different nickname, or identify via /msg NickServ identify <password>."
31 Aug 19:00:44 - GOT NOTICE from "freenode-connect": "Due to the persistent ongoing spam, all new connections are being set +R (block messages from unidentified users) and will be scanned for vulnerabilities. This will not harm your computer, and vulnerable hosts will be notified."
31 Aug 19:00:49 - Unhandled message: { prefix: 'weber.freenode.net',
  server: 'weber.freenode.net',
  command: '477',
  rawCommand: '477',
  commandType: 'normal',
  args:
   [ '<USERNAME>',
     '#channel',
     'Cannot join channel (+r) - you need to be identified with services - see https://freenode.net/kb/answer/registration' ] }
31 Aug 19:00:49 - GOT NOTICE from "NickServ": "You are now identified for <USERNAME>."
CyclingPingTimer 0: server silent for too long, let's send a PING
31 Aug 19:01:04 - SEND: PING 1
CyclingPingTimer 0: server silent for too long, let's send a PING
31 Aug 19:01:19 - SEND: PING 2
~~~

and the bot never joins the channel.
(i guess this is what the `autoRejoin:true` is supposed to handle; but afaict auto-rejoining only happens as a response to be `KICK`ed out of the channel)

umlaeute:
re-issueing a `JOIN` after the server responded with the "You are now identified for <USERNAME>." message fixes the problem.
However, I have no idea how to properly detect that message (short of doing string-matching)

michalrud:
Yeah, that would be because - to avoid such string matching, which I guess would be server-specific - the identify code assumes that next message that comes after sending the is information that it suceeded. Maybe it would be enough to check if that message came from NickServ?

michalrud:
@jwflory Not yet. Thanks for reminding though, I will try to work on it today

michalrud:
:warning: Please do not merge yet, as module tests for rejoining to the channel are still missing. I tried to use an [internal nodeirc joined channel list](https://node-irc.readthedocs.io/en/latest/API.html#Client.chans) and I'm not sure if it works - it would be ready to be tested in real connections, though.

jwflory:
@michalrud Thanks, let us know if you need any help.

Tjzabel:
@michalrud hey, how's this coming along?

jwflory:
@michalrud Thanks for following up. Could you file a new issue with some starter points when someone has a chance to look at the tests in the future? Once you file the issue, let's merge this in. :+1:

jwflory:
#76 covers the follow-up work for this PR. Let's merge this!! :clapper:

PR #70, opened 2018-09-30
We need to state outright that channels aren't supported. 

robbyoconnor:
Add note to the installation docs that Telegram channels are not supported

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -137,6 +137,9 @@ IRC
 Telegram
 ^^^^^^^^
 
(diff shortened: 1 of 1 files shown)


Tjzabel:
awesome, thanks for the rebase as well :+1:

robbyoconnor:
I always rebase

jwflory:
@robbyoconnor Thanks for the catch and thanks @Tjzabel for jumping on and reviewing here. :+1:

PR #72, opened 2018-10-06
IRC Handler Refactor

xforever1313:
This pull request splits each of the various IRC message handlers (msg, kick, join, part, etc) into separate classes instead of having one massive function in TeleIrc.js.

This not only makes it to unit test each operation, but it allows an easy way to expand the functionality of these commands.  For example, maybe at some point in the future we want a way for a user to configure the join/part/kick message.  Its easier to implement and unit test if these actions are in separate classes, instead of all in one massive function.

All of the new classes have a constructor and one function.  The constructor takes in information that is needed in order to construct the message to send to Telegram, and a callback.  The one function generates the message to send to telegram, and calls the callback that in passed in during construction.  In unit-test land, the passed in callback simply sets a local variable, while in production, the call-back enqueues the message to telegram.

Unit tests have been added as well.


jwflory:
@xforever1313 Thanks for making this PR. I think this makes the project much easier to work with. I hope @michalrud is able to get #65 in first before we restructure the entire project.

jwflory:
@xforever1313 Since we are waiting for @michalrud in #65, [here is some instant feedback](https://codeclimate.com/github/RITlug/teleirc/pull/72) on the pull request. CodeClimate caught a few cases of duplicated code that could be refactored. If you have time before #65 is merged, feel free to work off of the CodeClimate feedback.

I'll try to do a detailed review over the next few days.

Tjzabel:
@xforever1313 thanks for the PR! This will definitely help future development due to the modularity :+1: 
As far as the code review, besides CodeClimate throwing a fit, I'm going to leave this to @jwflory. I'm not familiar enough with JS to give a meaningful enough review.

xforever1313:
Removed the duplication for checking black-listed usernames... however the ones in the constructor I'd argue is a false negative.  The constructors are assigning member variables to the classes, its not something that can be "commonized", as the member variables in the classes are, well, separate :P.

lib/IrcHandlers/IrcActionHandler.js
```diff
@@ -0,0 +1,50 @@
...
+    // ---------------- Functions ----------------
+
+    /**
+     * Constructs the message and fires this class's callback.
+     * @param {string} username - The user who performed the action.
+     * @param {string} channel - The channel the user has performed the action in.
+     * @param {string} userAction - The action the user performed.
+     */
+    ReportAction(username, channel, userAction) {
+        if (this.Enabled == false ) {
```
thenaterhood: You've got an extra space before the closing paren here.

Also, can we use a type-aware comparison here? `===` is generally preferable if the types should match. If this can be `===` I'd suggest a quick run-through of the other classes and see if the same can be done there.

lib/IrcHandlers/IrcActionHandler.js
```diff
@@ -0,0 +1,50 @@
+const Helpers = require("../Helpers.js");
+
+/**
+ * Handles the event when a user performs an action in the IRC channel.
```
thenaterhood: By "performs an action" I assume you mean a `/me` command? For posterity, we might want to clarify that here, since "performing an action" is pretty ambiguous.

lib/IrcHandlers/IrcJoinHandler.js
```diff
@@ -0,0 +1,37 @@
...
+     * Constructs the message and fires this class's callback.
+     * @param {string} channel - The channel the user has joined.
+     * @param {string} username - The user that joined the channel.
+     */
+    ReportJoin(channel, username) {
+        if (this.Enabled == false ) {
+            return;
+        }
+
+        let message = username + " has joined " + channel + " channel.";
```
thenaterhood: I think you're missing a "the" in this sentence. Should this be "username has joined the $channel channel."?

lib/IrcHandlers/IrcMessageHandler.js
```diff
@@ -0,0 +1,49 @@
...
+
+    // ---------------- Functions ----------------
+
+    /**
+     * Constructs the message and fires this class's callback.
+     * @param {string} username - The user who sent the message.
+     * @param {string} channel - The channel the user has sent the message in.
+     * @param {string} userMessage - The message the user sent.
+     */
+    ReportMessage(username, channel, userMessage) {
```
thenaterhood: Would a better name for this method be `RelayMessage` or `ForwardMessage`?

lib/TeleIrc.js
```diff
@@ -497,6 +470,10 @@ class TeleIrc {
       message
     );
   }
+
+  status(message) {
```
thenaterhood: Is this a debugging relic?

xforever1313: It is not... typically when I have a class that reports logging, I have a single function called "Status" that handles how the logging takes place.  Maybe a timestamp wants to come before the message.  Maybe we want to do Console.error instead of Console.log.

I haven't converted all of the other console.log calls with this pull; I was planning on doing that when I attempt to make the Telegram side less of a cluster :P.

lib/IrcHandlers/IrcJoinHandler.js
xforever1313: I make it "$user has joined $channel" to be consistent with part messages ("$user has left $channel").

lib/IrcHandlers/IrcActionHandler.js
xforever1313: Added a comment.

lib/IrcHandlers/IrcActionHandler.js
xforever1313: Went through all of the new handler classes and made them === instead of ==.

lib/IrcHandlers/IrcMessageHandler.js
xforever1313: RelayMessage it is!

jwflory:
@thenaterhood Thanks for leaving a helpful review. :smile: 

@xforever1313 #65 is now merged. Could you please rebase your changes on top of the `HEAD` of master branch? I'm not sure how `git rebase` savvy you are, but there is an excellent [blog post guide](https://nathanleclaire.com/blog/2014/09/14/dont-be-scared-of-git-rebase/) and a [edX tutorial](https://github.com/edx/edx-platform/wiki/How-to-Rebase-a-Pull-Request) on rebasing that might help.

xforever1313:
Rebase is done, but one unit test is failing... and that's because the MockIrcBot can't handle multiple things subscribing to the same event... I'll look into it tomorrow or Friday.

xforever1313:
*phew* I think I got everything working again.  That was a rough rebase.

Had to add a new class to handle messages from NickServ.  Also fixed an issue where the bot would crash with NickServ enabled, but "nickservRegisteredNick" isn't in the config.js.  That setting isn't even in config.js, so I wonder if that is a bug?

All Unit Tests pass, and tested coms between IRC and Telegram, everything seems to work.

Issue #67, opened 2018-09-06
spurious disconnect


umlaeute:
this morning my teleirc bridge suddenly disconnected from the IRC.

the relevant logs from my docker-container running teleirc are:
~~~
Sep 06 07:33:20 docker1 env[9707]: teleirc_1  | Unhandled rejection TimeoutError: operation timed out
Sep 06 07:33:20 docker1 env[9707]: teleirc_1  |     at afterTimeout (/opt/teleirc/node_modules/request-promise/node_modules/bluebird/js/main/timers.js:16:15)
Sep 06 07:33:20 docker1 env[9707]: teleirc_1  |     at Timeout.timeoutTimeout [as _onTimeout] (/opt/teleirc/node_modules/request-promise/node_modules/bluebird/js/main/tim
Sep 06 07:33:20 docker1 env[9707]: teleirc_1  |     at ontimeout (timers.js:498:11)
Sep 06 07:33:20 docker1 env[9707]: teleirc_1  |     at tryOnTimeout (timers.js:323:5)
Sep 06 07:33:20 docker1 env[9707]: teleirc_1  |     at Timer.listOnTimeout (timers.js:290:5)
Sep 06 07:33:25 docker1 env[9707]: teleirc_1  | Unhandled rejection RequestError: Error: connect EADDRNOTAVAIL 2001:67c:4e8:f004::9:443 - Local (:::0)
Sep 06 07:33:25 docker1 env[9707]: teleirc_1  |     at new RequestError (/opt/teleirc/node_modules/request-promise/lib/errors.js:11:15)
Sep 06 07:33:25 docker1 env[9707]: teleirc_1  |     at Request.RP$callback [as _callback] (/opt/teleirc/node_modules/request-promise/lib/rp.js:60:32)
Sep 06 07:33:25 docker1 env[9707]: teleirc_1  |     at self.callback (/opt/teleirc/node_modules/request/request.js:186:22)
Sep 06 07:33:25 docker1 env[9707]: teleirc_1  |     at emitOne (events.js:116:13)
Sep 06 07:33:25 docker1 env[9707]: teleirc_1  |     at Request.emit (events.js:211:7)
Sep 06 07:33:25 docker1 env[9707]: teleirc_1  |     at Request.onRequestError (/opt/teleirc/node_modules/request/request.js:845:8)
Sep 06 07:33:25 docker1 env[9707]: teleirc_1  |     at emitOne (events.js:116:13)
[...]

PR #74, opened 2018-10-07
Crash on Telegram bot errors

### LINKED ISSUE #15: teleirc does not "crash" if it encounters a fatal error
teleirc does not crash and return an exit code if it encounters an unrecoverable error. Rather, it fails and hangs without exiting.

thenaterhood:
This may resolve #15 and #67 but since I can't replicate the exact scenario for those, I can't make guarantees.

**WARNING** this updates the version of the telegram bot library in use, which is a requirement for these changes to work.

### DIFF
--- a/teleirc.js
+++ b/teleirc.js
@@ -28,6 +28,25 @@ let ircbot = new irc.Client(config.irc.server, config.irc.botName, {
 // Create the telegram bot side with the settings specified in config object above
 console.log("Starting up bot on Telegram...");
 let tgbot = new tg(config.token, { polling: true });
+// This is super rudimentary error handling. If the Telegram bot
+// throws an error, the library we use simply logs it and appears
+// to hang. This may be a bug in that library (see
+// https://github.com/yagop/node-telegram-bot-api/issues/657).
+// This code forces teleirc to fully crash when it encounters a polling
+// error so that its parent process/container can restart it.
+// It may be possible to handle this in a nicer way, but issues
+// open on the bot library indicate that things may not be working right
+// as of October 2018.
+tgbot.on('polling_error', (error) => {
+    console.log("Fatal Telegram polling error: " + error.code);
+    console.log(error.stack);
+    process.exit(1);
+});
+tgbot.on('webhook_error', (error) => {
+    console.log("Fatal Telegram webhook error: " + error.code);
+    console.log(error.stack);
+    process.exit(1);
+});
 
 teleIrc.initStage3_initBots(ircbot, tgbot);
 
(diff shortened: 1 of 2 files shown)


Tjzabel:
Hm, CodeClimate has been super pedantic lately. @thenaterhood thanks for the PR! @jwflory what do you think?

Tjzabel:
I think we should keep this PR open until [PR #72](https://github.com/RITlug/teleirc/pull/72) goes through.
The location of the error catching may need to be changed depending on this PR going through.

thenaterhood:
AFAIK #72 only touches the IRC side not the Telegram side. Once the Telegram side gets a similar refactor then the error handling code could move.

Code climate has configurable thresholds. We could decide on what thresholds are reasonable for teleirc and add a config file to make certain checks less pedantic. The defaults are here: https://docs.codeclimate.com/docs/advanced-configuration#section-default-checks

Tjzabel:
@thenaterhood it seems #72 does only touch the IRC side. There are still many items taken out of the TeleIrc.js file, but I suppose this won't affect anything with this PR.

And yes, I do think after seeing the past few PRs, that it may be a good idea to dial down CodeClimate's pedantic level.

xforever1313:
Changes look fine to me.  They shouldn't interfere with #72, as this only changes teleirc.js, while #72 doesn't even touch that file.

jwflory:
@thenaterhood #65 is now merged. Could you please rebase your changes on top of the `HEAD` of master branch? I'm not sure how `git rebase` savvy you are, but there is an excellent [blog post guide](https://nathanleclaire.com/blog/2014/09/14/dont-be-scared-of-git-rebase/) and a [edX tutorial](https://github.com/edx/edx-platform/wiki/How-to-Rebase-a-Pull-Request) on rebasing that might help.

Alternatively, we could wait for #72 to get merged before rebasing, but I see @xforever1313 noting that the changes are independent of each other.

thenaterhood:
Rebased :fireworks:

teleirc.js
```diff
@@ -28,6 +28,25 @@ let ircbot = new irc.Client(config.irc.server, config.irc.botName, {
...
 let tgbot = new tg(config.token, { polling: true });
+// This is super rudimentary error handling. If the Telegram bot
+// throws an error, the library we use simply logs it and appears
+// to hang. This may be a bug in that library (see
+// https://github.com/yagop/node-telegram-bot-api/issues/657).
+// This code forces teleirc to fully crash when it encounters a polling
+// error so that its parent process/container can restart it.
+// It may be possible to handle this in a nicer way, but issues
+// open on the bot library indicate that things may not be working right
+// as of October 2018.
```
jwflory: Thanks for the context explanation here. :+1:

PR #80, opened 2018-10-15
Telegram Refactor

xforever1313:
So, this part two of the refactor I've been working on to make the code more, uh, sane.

This pull splits up the massive handleTelegramMsg into separate classes - one for each type of thing we support Telegram side.

Basic design is this; it is the same as the IRC refactor:

There is a new class called "TgEventListener".  The Telegram library we use does not have an join/part/message event like the IRC library we use does.  This class emulates that.  In TeleIrc.js, we subscribe to these events.

Each event has its own event handler.  Sticker, Photo, Part (leave), Message, Join, and Document.  Photo is a base class, which by default posts the URL of the photo to the IRC channel.  There is an imgur version of this class which will upload to imgur instead (imgur version inherits and overrides the default class).  While Photo is currently the only class that has this polymorphism, there is nothing stopping something like the Document handler from doing the same thing at some point in the future.

Added several unit tests to test each handler.

I tested everything (sticker, file, photo, message, join, part)... except for uploading to imgur (I don't have API keys); though I don't see any reason why it wouldn't work (famous last words :P).

And with this, we can wave the scary handleTelegramMsg function goodbye! :D


xforever1313:
...I'm not entirely sure why my pull included ALL of the commit history including the IRC handler stuff... odd...

lib/TeleIrc.js
```diff
@@ -298,114 +391,6 @@ class TeleIrc {
     this.tgRateLimiter.queueMessage(messageString);
   }
 
-  /**
-   * Handles a message from telegram.
-   */
-  handleTelegramMsg(msg) {
```
xforever1313: Good Riddance!

lib/TelegramHandlers/TgPhotoHandler.js
```diff
@@ -0,0 +1,113 @@
...
+
+    // ---------------- Functions ----------------
+
+    /**
+     * 
+     * @param {*} from - Object that contains the information about the user name.
+     * @param {*} photoList - The information about the photos we can get from Telegram.
+     * @param {string} - The caption of the photo, if any.
+     */
+    async ReportPhoto(from, photoList, caption) {
```
thenaterhood: How about `RelayPhoto` or `SendPhoto`? `ReportPhoto` makes me think of a moderation-type function like `Report Abuse`.

lib/TelegramHandlers/TgPartHandler.js
```diff
@@ -0,0 +1,41 @@
...
+    }
+
+    // ---------------- Functions ----------------
+
+    /**
+     * Constructs the message and fires this class's callback.
+     * @param {*} partedUser - Object that contains the information about the user name.
+     */
+    ReportPart(partedUser) {
+        if (this.Enabled === false) {
```
thenaterhood: If this is generally intended to be a boolean, why not just `if (this.Enabled) {`?

lib/TelegramHandlers/TgPhotoHandler.js
xforever1313: All better.

lib/TelegramHandlers/TgPartHandler.js
xforever1313: All better.

Issue #81, opened 2018-10-15
Bot should check API responsiveness upon startup


ponkyh:
Had an issue when bot started just fine and joined the IRC channel but nothing got redirected from IRC to TG or vice versa. No errors. Tried to curl the api but got "curl: (35) Unknown SSL protocol error in connection to api.telegram.org:443", curling from the same host with --interface "different ip" worked, so I'd guess it's a some kind of rate limit/blacklist on Telegrams end.

Maybe the bot could do a /getMe API-call upon startup to check that API is actually responding? getMe response looks like this:

{"ok":true,"result":{"id":xxx,"is_bot":true,"first_name":"botnick","username":"botname"}}

Tjzabel:
@jwflory @thenaterhood I'm assuming since the PR referencing this issue has been merged, that this issue has been fixed?

thenaterhood:
In theory yeah

PR #83, opened 2018-10-23
Check if we can actually reach Telegram on start

thenaterhood:
This implements the suggestion in #81. I think this is slightly redundant with the change that makes teleirc crash on Telegram polling errors, but I haven't figured out how to make Telegram rate limit me in a way that I can cause this (nor do I want to figure out how to get myself rate limited).

If this saves some debugging time for someone in the future, then it's worthwhile.

### DIFF
--- a/lib/TeleIrc.js
+++ b/lib/TeleIrc.js
@@ -210,8 +210,7 @@ class TeleIrc {
 
     // Verifies that we can, in fact, reach the Telegram API and
     // are not being rate limited.
-    let getmePromise = this.tgbot.getMe();
-    getmePromise.then(function(result) {
+    this.tgbot.getMe().then(function(result) {
         console.log("Connected to Telegram. Our bot name is " + result.username);
     })
     .catch(function(result) {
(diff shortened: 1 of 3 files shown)


thenaterhood:
Ah wouldja look at that. I should know better than to move code around. I'll fix that test tomorrow :)

lib/TeleIrc.js
```diff
@@ -188,6 +188,35 @@ class TeleIrc {
...
+        process.exit(1);
+    });
+    this.tgbot.on('webhook_error', (error) => {
+        console.log("Fatal Telegram webhook error: " + error.code);
+        console.log(error.stack);
+        process.exit(1);
+    });
+
+    // Verifies that we can, in fact, reach the Telegram API and
+// are not being rate limited.
```
xforever1313: A comment isn't indented

lib/TeleIrc.js
```diff
@@ -188,6 +188,35 @@ class TeleIrc {
...
+    let getmePromise = this.tgbot.getMe();
```
xforever1313: Could use async/await here... though since this is only one promise, and not a giant cascade of them (like the imgur stuff was), I'm okay with leaving it as is.

PR #89, opened 2018-11-24
Fix Fedora container, validate on Fedora 29 (closes #86)

### LINKED ISSUE #86: Fix Fedora container image
The Fedora container image fails to start. Need to get it working and running as expected again. Also, upgrade to Fedora 29?

jwflory:
This one wasn't too complicated, but took some playing around. What changed:

* Fix statements using old paths for files that no longer exist there
* Don't uninstall `make`, since it uninstalls `nodejs`
* `yarn` emits an error code, even if it actually was successful; therefore, run the next command regardless if it fails

### DIFF
--- a/containers/Dockerfile.fedora
+++ b/containers/Dockerfile.fedora
@@ -1,19 +1,21 @@
-FROM fedora
+FROM fedora:29
+
 # Cache layers which will not change
 RUN groupadd teleirc -g 65532 \
-&& useradd -u 65532 -g teleirc -s /bin/bash -M -d /opt/teleirc teleirc
+    && useradd -u 65532 -g teleirc -s /bin/bash -M -d /opt/teleirc teleirc
 
 WORKDIR /opt/teleirc
-CMD ["node", "teleirc.js"]
 
 COPY . /opt/teleirc/
-COPY config.js.example /opt/teleirc/config.js
+COPY config.js /opt/teleirc/config.js
 
 RUN curl --silent --location https://dl.yarnpkg.com/rpm/yarn.repo | tee /etc/yum.repos.d/yarn.repo \
-&& dnf -y update --setopt=deltarpm=false \
-&& dnf clean all \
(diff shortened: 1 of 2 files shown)


containers/Dockerfile.fedora
```diff
@@ -1,19 +1,21 @@
 FROM fedora
```
Tjzabel: It may be a good idea to specify `fedora:29` instead of leaving it at `fedora:latest`. This way, we know exactly which fedora version is being used. This makes more sense with Fedora's 6-month release cycles.

This then allows for time to make sure the next fedora version docker image works fine.

containers/Dockerfile.fedora
```diff
@@ -1,19 +1,21 @@
...
-&& dnf clean all \
-&& dnf -y install nodejs libicu-devel python gcc-c++ make yarn --setopt=deltarpm=false \
-&& yarn \
-&& dnf -y remove libicu-devel gcc-c++ make \
-&& dnf clean all \
-&& chown -R teleirc:teleirc /opt/teleirc
+    && dnf -y upgrade --setopt=deltarpm=false \
+    && dnf -y install nodejs libicu-devel python gcc-c++ make yarn --setopt=deltarpm=false \
+    && yarn \
+    ; dnf -y remove libicu-devel gcc-c++ \
```
Tjzabel: what's the reasoning for installing `gcc-c++` and then removing it in the next line? Also, why does this line start with a semicolon instead of continuing with the `&&`?

containers/Dockerfile.fedora
jwflory: Good idea. I'll pin to Fedora 29.

containers/Dockerfile.fedora
Tjzabel: @jwflory ah, makes sense now. That message didn't make that portion clear, but now it is.

PR #90, opened 2018-11-24
Remove/update calls to retired method names (closes #87, closes #78)

jwflory:
After #80, a few unused method calls were lying around. For the ones where a clear rename existed, I updated the method name. For the one part where I could not, I removed it.

I manually validated that the bot successfully started and ran. I also [verified](https://i.imgur.com/EShsvZf.jpg) Imgur image upload is working.

PR #91, opened 2018-11-24
Set NTBA_FIX_319=1 in env.example (closes #88)

### LINKED ISSUE #88: Support cancellation of promises with node-telegram-bot-api
Every time the bot runs, this text blurb pops up:

```
node-telegram-bot-api deprecated Automatic enabling of cancellation of promises is deprecated.
In the future, you will have to enable it yourself.
See https://github.com/yagop/node-telegram-bot-api/issues/319. module.js:653:30
```

It would be nice to ensure promises can be canceled, since it seems we're using them actively. I'll try to dig into this one and see if I can find a fix.

jwflory:
This resolves a deprecation warning given by node-telegram-bot-api about the "automatic enabling of cancellation of promises". Given context from the [issue linked in the upstream project](https://github.com/yagop/node-telegram-bot-api/issues/319#issuecomment-324963294), setting this environment variable will ensure backwards-compatibility with future releases. If we need this feature, then we need to enable it when we need it.

### DIFF
--- a/env.example
+++ b/env.example
@@ -41,3 +41,13 @@ TELEGRAM_CHAT_ID=-0000000000000
 USE_IMGUR_FOR_IMAGES=false
 IMGUR_CLIENT_ID=0000000000
 
+
+###############################################################################
+#                                                                             #
+#                      Misc. required env variables                           #
+#                -- not meant for user config, don't touch --                 #
+#                                                                             #
+###############################################################################
+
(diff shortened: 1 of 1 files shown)


jwflory:
CI could be failing because of #90. I can rebase this PR after #90 is merged to see if CI tests complete successfully.

Issue #77, opened 2018-10-12
Create doc page on development environment


jwflory:
Need better developer docs for new contributors. Working with @icflournoy @zedjones and noticed the docs aren't clear in this way.

jwflory:
_Discussed in 2019-02-09 RITlug developers' meeting._

---

Briefly discussed. @Tjzabel and @nic-hartley both are using a test Telegram group and test IRC channel for a dev environment. New docs should walk through this process to set up a dev environment to test changes (different kind of testing than CI tests).

jwflory:
PR #133 will close this issue.

PR #92, opened 2018-11-24
Use pipenv for docs development environment

jwflory:
To make it easier for people to contribute to the docs, I added a Pipfile to support using [pipenv](https://pipenv.readthedocs.io/). It takes the mess out of virtualenvs in Python (especially helpful since this is a Node project, and new developers might not be familiar with Python tool chains).

I'll support this with how-to documentation, as per #77.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -51,19 +51,6 @@ node_modules
 
 # ----------------------------- Python (for docs) ------------------------------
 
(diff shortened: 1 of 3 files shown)


robbyoconnor:
Shouldn't we update docs for working this? That said, this can probably be done in a follow-up PR

PR #102, opened 2018-12-05
Splitting of messages sent to IRC, prepending TG username to each one.

### LINKED ISSUE #53: multi-line messages should have each lined prefixed with the userid
Currently, when a message with newlines is posted to Telegram, it will be automatically splitted by the bot (well, technically by the underlying irc client library), so when `user1` posts a message like this to Telegram (notice the newline):

```
This is my message.
<user2> I am not very smart
```

teleirc will post following messages to the IRC channel:

```
<teleirc>: <user1> This is my message.
<teleirc>: <user2> I am not very smart
```

Therefore, misleading other IRC users into thinking that `user2` admitted his/her own shortcomings.

I will be preparing a PR with a fix. The code is mostly ready, but I would like to test it more before I would consider it ready _enough_ as I didn't actually test it with telegram nor IRC.

michalrud:
Should fix #53. Not tested in live environment.

Also fixed / improved some UTs. Especially hardcoded some config values in tests - some of them would fail if you don't have teleirc configured properly in the first place, what should not be required in UT IMHO.

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -142,6 +142,11 @@ IRC
 - **IRC_NICKSERV_PASS**:
     - IRC password for your bot to use in order to complete IRC authentication (default: ``""``)
 
+- **IRC_MAX_MESSAGE_LENGTH**:
+  
+    - Maximum length of the message that can be sent to IRC. Longer messages
+      will be splitted into multiple messages. (default: ``400``)
+
 Telegram
 ^^^^^^^^
 
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -143,9 +143,8 @@ IRC
     - IRC password for your bot to use in order to complete IRC authentication (default: ``""``)
 
 - **IRC_MAX_MESSAGE_LENGTH**:
-  
     - Maximum length of the message that can be sent to IRC. Longer messages
-      will be splitted into multiple messages. (default: ``400``)
+      will be split into multiple messages. (default: ``400``)
 
 Telegram
 ^^^^^^^^
--- a/lib/TeleIrc.js
+++ b/lib/TeleIrc.js
@@ -41,6 +41,7 @@ class TeleIrc {
     this.defaultIrcSuffix = "";
     this.defaultShowJoinMessage = false;
     this.defaultShowLeaveMessage = false;
+    this.defaultMaxMessageLength = 400;
   }
 
   // ---------------- Functions ----------------
@@ -110,6 +111,12 @@ class TeleIrc {
         this.config.irc.showLeaveMessage,
         this.defaultShowLeaveMessage,
         "showLeaveMessage"
+      );
+
+      this.config.irc.maxMessageLength = this.checkConfigOption(
+        this.config.irc.maxMessageLength,
+        this.defaultMaxMessageLength,
+        "maxMessageLength"
       );
     } else {
       throw new TeleIrcException(
@@ -445,11 +452,22 @@ class TeleIrc {
    * Sends a message to the IRC channel.
    * @param {String} message - The message to send to the IRC channel.
    */
-  sendMessageToIrc(message) {
-    this.ircbot.say(
-      this.config.irc.channel,
-      message
-    );
+  sendMessageToIrc(from, message) {
+    const ircbot = this.ircbot;
+    const messagePrefix = this.config.irc.prefix + from + this.config.irc.suffix + " ";
+    const channel = this.config.irc.channel;
+    const maxLength = this.config.irc.maxMessageLength;
+
+    message.toString().split(/\r?\n/).filter(function(line) {
+      return line.length > 0;
+    }).forEach(function(line) {
+      var linesToSend = ircbot._splitLongLines(line, maxLength - messagePrefix.length, []);
+      
+      linesToSend.forEach(function(toSend) {
+        ircbot.say(channel, messagePrefix + toSend);
+      });
+    });
+    
   }
 
   /**
--- a/lib/TeleIrc.js
+++ b/lib/TeleIrc.js
@@ -450,38 +450,13 @@ class TeleIrc {
 
   /**
    * Sends a message to the IRC channel.
-   * @param {(String|Object)} input - The message to send to the IRC channel.
-   * @param {String} input.from - Name of the user from which the message was received.
-   * @param {String} input.message - Actual message to be sent.
-   * 
-   * If `input` is provided as string, message is forwarded as-is. Otherwise, a prefix is
-   * generated. Messages longer than config.irc.maxMessageLength will be splitted into
-   * multiple messages.
+   * @param {String} message - The message to send to the IRC channel.
    */
-  sendMessageToIrc(input) {
-    var prefix = "";
-    var message = "";
-    if (typeof input === "string") {
-      message = input;
-    }
-    else {
-      prefix = this.config.irc.prefix + input.from + this.config.irc.suffix + " ";
-      message = input.message;
-    }
-    const ircbot = this.ircbot;
-    const channel = this.config.irc.channel;
-    const maxLength = this.config.irc.maxMessageLength;
-
-    message.toString().split(/\r?\n/).filter(function(line) {
-      return line.length > 0;
-    }).forEach(function(line) {
-      var linesToSend = ircbot._splitLongLines(line, maxLength - prefix.length, []);
-      
-      linesToSend.forEach(function(toSend) {
-        ircbot.say(channel, prefix + toSend);
-      });
-    });
-    
+  sendMessageToIrc(message) {
+    this.ircbot.say(
+      this.config.irc.channel,
+      message
+    );
   }
 
   /**
--- a/lib/TeleIrc.js
+++ b/lib/TeleIrc.js
@@ -419,7 +419,7 @@ class TeleIrc {
 
   /**
    * Sends a message to telegram.
-   * @param messageString - The message to send.
+   * @param {String} messageString - The message to send.
    */
   queueTelegramMessage(messageString) {
     this.tgRateLimiter.queueMessage(messageString);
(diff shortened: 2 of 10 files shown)


Tjzabel:
@michalrud first off, thanks for the PR!

I've run a few tests off your branch. Firstly, multiline messages now work! Great work on this! :tada: :tada: 
However, telegram photos don't go through to IRC, and so imgur uploading doesn't work.

docs/installation.rst
```diff
@@ -134,6 +134,11 @@ IRC
 
     - IRC server you wish to connect to (default: ``chat.freenode.net``)
 
+- **IRC_MAX_MESSAGE_LENGTH**:
```
Tjzabel: If you are going to add this to the available configs, you'll need to add `IRC_MAX_MESSAGE_LENGTH` to the available options in the `env.example` file.

lib/TeleIrc.js
```diff
@@ -442,11 +449,22 @@ class TeleIrc {
...
-      this.config.irc.channel,
-      message
-    );
+  sendMessageToIrc(from, message) {
+    const ircbot = this.ircbot;
+    const messagePrefix = this.config.irc.prefix + from + this.config.irc.suffix + " ";
+    const channel = this.config.irc.channel;
+    const maxLength = this.config.irc.maxMessageLength;
+
+    message.toString().split(/\r?\n/).filter(function(line) {
```
Tjzabel: Looks like the `toString()` method is failing when the `message` is an image, which naturally wouldn't have a `toString` method.

lib/TeleIrc.js
```diff
@@ -442,11 +449,22 @@ class TeleIrc {
...
+      return line.length > 0;
+    }).forEach(function(line) {
+      var linesToSend = ircbot._splitLongLines(line, maxLength - messagePrefix.length, []);
```
Tjzabel: Instead of using this methodology to split the messages based on the created `maxLength`, it may be useful to look into the `messageSplit` option available through the [irc API](https://node-irc.readthedocs.io/en/latest/API.html#client).

Tjzabel:
I've given my review. Since the changes I've requested are relatively large changes to how you've done things, I haven't taken much of a look to the unit tests you've created. Once the core functionality is working, I will take my time to go through the tests.

Thanks again :smile:

michalrud:
> @michalrud first off, thanks for the PR!
> ## At first glance
> 
> I've run a few tests off your branch. Firstly, multiline messages now work! Great work on this! tada tada
> However, telegram photos don't go through to IRC, and so imgur photo uploading doesn't work.
> ## A little more in-depth
> 
> Here is a stack trace of what happens when photos get uploaded to Telegram:
> 
> ```
> (node:4058) UnhandledPromiseRejectionWarning: TypeError: Cannot read property 'toString' of undefined
>     at TeleIrc.sendMessageToIrc (/home/tim/test/teleirc/lib/TeleIrc.js:458:13)
>     at TgImgurPhotoHandler.RelayPhotoMessage (/home/tim/test/teleirc/lib/TelegramHandlers/TgPhotoHandler.js:54:18)
>     at process._tickCallback (internal/process/next_tick.js:68:7)
> (node:4058) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 1)
> (node:4058) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
> (node:4058) UnhandledPromiseRejectionWarning: TypeError: Cannot read property 'toString' of undefined
>     at TeleIrc.sendMessageToIrc (/home/tim/test/teleirc/lib/TeleIrc.js:458:13)
>     at TgImgurPhotoHandler.RelayPhotoMessage (/home/tim/test/teleirc/lib/TelegramHandlers/TgPhotoHandler.js:54:18)
>     at process._tickCallback (internal/process/next_tick.js:68:7)
> (node:4058) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 2)
> ```

Okay - I know exactly what might have caused it. I'll fix it tomorrow + add an appropriate test. In short - now sendMessageToIrc receives two parameters, `(sender, message)`, while it previously just received message - as `sendMessageToIrc` now prepends the prefix by itself.

lib/TeleIrc.js
michalrud: I've considered it at first, but then it won't prepend the prefix to the next lines, and that's what we want to avoid :wink:

michalrud:
I think I've fixed the issue in the simplest way possible, even if I don't like it that much. IMHO a cleaner way would be to expect all `action` callbacks to communicate using structs, but maybe I'm too broken by strongly typed languages :) Also I was thinking about providing a struct interface compatibility to `queueTelegramMessage`, but to be honest that wouldn't be used currently, so why add more unnecessary complication?

If you prefer, though, I can rework all `Tg*Handler` pieces to use the struct interface. It will IMHO make the code cleaner, but also the change bigger. You decide :)

Tjzabel:
@michalrud thanks for the updates! I will look through the updates over the weekend. I will defer the opinion of adding structs to @xforever1313 and @thenaterhood since they would be better able to answer that question.

Tjzabel:
Also, the updated tests look good to me :+1:

jwflory:
@michalrud Thanks again for your time and efforts on this PR! It is super appreciated. :tada:

lib/TeleIrc.js
```diff
@@ -440,13 +447,38 @@ class TeleIrc {
 
   /**
    * Sends a message to the IRC channel.
-   * @param {String} message - The message to send to the IRC channel.
+   * @param {(String|Object)} input - The message to send to the IRC channel.
+   * @param {String} input.from - Name of the user from which the message was received.
+   * @param {String} input.message - Actual message to be sent.
+   * 
+   * If `input` is provided as string, message is forwarded as-is. Otherwise, a prefix is
```
thenaterhood: I'm not sure this is the place for this logic. It would fit better in https://github.com/RITlug/teleirc/blob/master/lib/TelegramHandlers/TgMessageHandler.js#L33, which already has all the data handy for assembling a message. The message splitting logic you're adding looks like it can pretty much just be moved to the TgMessageHandler, swapping out the `ircbot.say` call for `this._action`. Moving it would also save a little work since you then don't need to worry about other data types.

As-is, this looks like a separation of concerns problem. There are two reasons this caught my attention:
- This is a method that originally took a prebuilt piece of data and passed it along to somewhere else. Here, it gained the need to know more things and is now doing work that was originally done elsewhere. Design-wise for teleirc, message formatting should generally go in the TelegramHandlers (which would format to send to IRC) or in the IrcHandlers (which would format to send to Telegram).
- A method that gains a need to handle an additional data type in the same parameter makes me raise an eyebrow. There can be valid reasons for this to happen (extensions to a legacy API, for example) but it's often a sign of a larger design problem.

Tjzabel:
@michalrud hey, how are the changes coming along?

jwflory:
@michalrud Hi! Did you get a chance to revisit this PR?

lib/TeleIrc.js
michalrud: I don't really agree:

> Design-wise for teleirc, message formatting should generally go in the TelegramHandlers (which would format to send to IRC)

The fact, that messages have certain length limitations is the issue on the IRC side. Why should Telegram part read IRC configuration options? What if you'd like to switch IRC backend to a different one, which doesn't have that limitation or has a different value?

> A method that gains a need to handle an additional data type in the same parameter makes me raise an eyebrow.

True - I left this as is to keep the change small and easily reviewable, for future refactoring.

But, in the end, it's your codebase, I will do the requested changes and you'll decide if you prefer my old approach or not :)

lib/TelegramHandlers/TgMessageHandler.js
```diff
@@ -31,14 +31,23 @@ class TgMessageHandler {
...
         }
 
         let username = TgHelpers.ResolveUserName(from);
-        let message = this._prefixSuffixConfig.prefix + username + this._prefixSuffixConfig.suffix + " " + userMessage;
-
-        this._action(message);
+        const messagePrefix = self._ircConfig.prefix + username + self._ircConfig.suffix + " ";
+        const messageSplitRegex = new RegExp(`.\{1,${self._ircConfig.maxMessageLength - messagePrefix.length}}`, 'g');
+
+        userMessage.toString().split(/\r?\n/).filter(function(line) {
```
xforever1313: Can you add a comment here just explaining that you are splitting the string?  If someone new looks at this code, it may not be obvious what is going on.

lib/TeleIrc.js
xforever1313: Heh, the problem with dynamic languages.

I wouldn't think of it as passing in an IRC Config object, but passing in an object that implements an interface that contains the prefix, suffix, and maximum length; an interface that IRC Config just so happens in implement.  A theoretical SlackConfig object or a WhatsAppConfig object could implement the same "interface" and be passed in instead.  Its why the parameter in the constructor was originally called "prefixSuffixConfig" instead of "ircConfig".

xforever1313:
I have one request, and that is just adding a comment.  Everything thing else looks good.

Once that comment is added: Approved!

lib/TelegramHandlers/TgMessageHandler.js
Tjzabel: @xforever1313 is a comment super necessary here? It *is* using the `.split` method, so the string splitting should be inherently known?

Tjzabel:
As an update, I've tested these past new commits locally, and everything works as planned :+1:

jwflory:
Excellent. I'm going to merge this one in. Thanks @michalrud for contributing this feature! :tada: 

Merging. :clapper:

PR #108, opened 2018-12-13
Rename containers to images since it's more representative

robbyoconnor:
This is a small nitpick since [containers aren't a thing](https://blog.jessfraz.com/post/containers-zones-jails-vms/), but images are.


robbyoconnor:
> I thought it would be cool to note, I shared this blog post with a Sys. Admin professor at RIT, and it's going to be part of required reading for students in future semesters.

This is awesome -- it should be required reading -- also her talk. "Containers" are really a collection of primitives.

Issue #41, opened 2017-10-11
Sending an image with a caption in telegram then editing the caption causes the image to get sent again


Chris-Bitler:
If you edit the caption on an image on telegram it will re-upload the image and send it to the irc chat again after the caption is edited:
![image](https://user-images.githubusercontent.com/11822578/31460252-c07a3a78-ae93-11e7-870a-83cb57699a19.png)
![image](https://user-images.githubusercontent.com/11822578/31460263-cd14302c-ae93-11e7-8839-ee051c93549c.png)

Chris-Bitler:
@jflory7 @robbyoconnor Is this intended or something we should fix? It might make more sense to send a message saying that the title was updated without reuploading the image.

robbyoconnor:
shrug

jflory7:
@Chris-Bitler: This behavior comes from the functionality either @thenaterhood or @repkam09 added for edited messages. Normally for a message, if it's edited, a new copy is sent as a new line to IRC. The edited message functionality is just resending the message, which in this case, is a copy of the image.

I'm more curious if we can retrieve the subtitle / caption and insert that for an image. If we could include the caption, then I'd actually prefer for it to resend the full message with the picture. Anyone know if this is possible?

robbyoconnor:
I won't have time to look at this.  I kind of want to rewrite this...

Chris-Bitler:
@jflory7 It is possible to get the caption and it actually already does, that's why it resends the image. Look at the title above the image in irc in the single quotes ('). It was @ dan first, then I edited it to be @ schneidy, which updated the title on the image. Also, in this case it's not just resending the image, it is completely re-uploading it to imgur (which means it's going through the entire code flow for photo messages again). That not be an issue, but it doesn't seem intended

jflory7:
Looking back on this, is there a way to distinguish regular messages versus edit messages? If we can fistinguish messages that are edits, I think we can figure out a way to extract only a caption from an edit message.

If not, I think we should close issue out as `wontfix`, since I think this will be intended behavior (however inconvenient).

xforever1313:
A potential approach:

Now that we have the TgHandlers split out, we could add a new function to each of those classes.  At the moment those classes only have the constructor and the RelayMessage function. We could add an additional function called RelayEditedMessage that we can hook into TgEventListener (and by extension, add "edited" events).

[...]

PR #110, opened 2018-12-27
Add [EDITED] prefix for edited messages on irc.

Palid:
This pull request adds edited prefix on IRC when the message was edited.

### DIFF
--- a/config.js
+++ b/config.js
@@ -14,6 +14,7 @@ let settings = {
     showLeaveMessage: process.env.IRC_SHOW_LEAVE_MESSAGE === "true" || true,
     nickservPassword: process.env.IRC_NICKSERV_PASS || "",
     nickservService: process.env.IRC_NICKSERV_SERVICE || "",
+    editedPrefix: process.env.IRC_EDITED_PREFIX || "[EDIT] ",
   },
   tg: {
     chatId: process.env.TELEGRAM_CHAT_ID || "-000000000",
--- a/env.example
+++ b/env.example
@@ -15,6 +15,7 @@ IRC_SUFFIX=">"
 IRC_SERVER=chat.freenode.net
 IRC_NICKSERV_SERVICE=NickServ
 IRC_NICKSERV_PASS=""
+IRC_EDITED_PREFIX="[EDIT] "
 
 
 ###############################################################################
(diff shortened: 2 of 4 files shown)


Tjzabel:
@Palid I've tested this locally off your branch, and everything works! This is definitely something we want to include :+1: 

This may also indirectly give some leeway to issue #41 

The *only* thing I want input on from others is whether or not `[EDIT]` is too strong of a pre-text message.

xforever1313:
I think this is a good first step to get the feature working; the change is simple and should work fine.  [EDIT] is fine in my opinion as a default prefix, especially since it is user configurable.  I don't have a  strong opinion either way though.

I see no reason why we can't merge this in to get this feature up and running.  However, I do have a concern going forward with the changes to TgEventListener.  TgEventListener shouldn't do any message manipulation, it should just determine which event to trigger, and let the TgEventHandlers handle the event.  See my comment in #41 (https://github.com/RITlug/teleirc/issues/41#issuecomment-450276016) for an idea to use this pattern.  That, however, is a lot more work to get up and running correctly, and a task for a different day (probably when we try to fix #41).

The other thing to check before merging is what if someone sets IRC_EDITED_PREFIX in the env file to ```null``` or ```undefined```?  Do we crash?  If not, then I approve this pull.  If it does, then checks should be added to TeleIrc.js's initStage2_telegramConfigValidation().

Thanks for the PR!

Tjzabel:
@xforever1313 makes sense. In terms of setting `IRC_EDITED_PREFIX` to null or undefined, nothing happens. If nothing is there, it just reverts to the default set as `[EDIT] `. 

I've tested this with literal `null, `undefined`, and nothing, in addition to random characters. Everything just gets passed as string literals, or the default otherwise.

Tjzabel:
@xforever1313 Since you think the merge is good to go, I'm going to go ahead and do the merge :+1: 

Feel free to either add a comment to update the previous issue, or open up another issue with your concerns for the future state of `TgEventListener` and `TgEventHandler`.

jwflory:
@Palid Thanks for submitting this feature! :tada:

Palid:
If nobody will refactor it before half of February I'll gladly do it. Unfortunately before that that I'm flooded with work and won't have enough time to properly refactor this.

PR #111, opened 2018-12-27
Proof of concept response to quoted data

Palid:
A proof of concept pull request to kind of properly show quoted/response to messages. Due to the refactor it's a bit problematic to do this without race conditions - see for instance `TgPhotoHandler.js` and method `RelayPhotoMessage`, that `await`s for file resolution. Unfortunately current mechanism has no way to properly await for the file resolution making my solution pretty unusable - to actually make it work properly codebase would require a bit of refactor.

### DIFF
--- a/config.js
+++ b/config.js
@@ -15,6 +15,7 @@ let settings = {
     nickservPassword: process.env.IRC_NICKSERV_PASS || "",
     nickservService: process.env.IRC_NICKSERV_SERVICE || "",
     editedPrefix: process.env.IRC_EDITED_PREFIX || "[EDIT] ",
+    replyToPrefix: process.env.IRC_REPLY_TO_PREFIX || "[REPLY TO THIS BELOW]",
   },
   tg: {
     chatId: process.env.TELEGRAM_CHAT_ID || "-000000000",
--- a/env.example
+++ b/env.example
@@ -16,7 +16,7 @@ IRC_SERVER=chat.freenode.net
 IRC_NICKSERV_SERVICE=NickServ
 IRC_NICKSERV_PASS=""
 IRC_EDITED_PREFIX="[EDIT] "
-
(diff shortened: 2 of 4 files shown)


xforever1313:
Thanks for the PR!  Unfortunately, I'm not comfortable merging this in its current state.

There's way too much logic inside of TgEventListener.  I was okay with doing it in the short-term for #110 because its a small and simple change.  This, however, is not.

TgEventListener should not do any message manipulation.  Its purpose is to determine which type of message we are and trigger an event.  It is up to the various TgEventHandler classes to then take the information and generate the message to send to IRC.  

A better approach would be for the TgEventListener to generate more events (e.g. replied_message, replied_photo, etc).  Then, all the TgEventHandler classes currently have a RelayMessage function.  There's nothing stopping us from adding a new RelayReplyMessage function in each class that takes in the message and the reply information to generate a message to send to IRC.  This is similar pattern to what I describe here in https://github.com/RITlug/teleirc/issues/41#issuecomment-450276016 . It also keeps the separation of concerns between each TgEventHandler class, and not in one spot in TgEventListener.

Palid:
@xforever1313 I'll try to take care of this in the middle of February, as I'm currently swarmed with work.
This pull request was more of a proof of concept to get your ideas of it. It kinda works, but is really meh.

jwflory:
Hi @Palid – are you still interested to work on this? We're not sure how to handle this PR since it is a proof-of-concept. I suggest opening a new PR when you have time to work on changes. But if you are confident you will have time to work on it this month, we can leave it open. What do you think?

If we don't hear back by the next Rochester developers' meeting on Saturday, Feb. 9th, we'll close this PR for now.

jwflory:
Hi @Palid, we're going to close this for now since it's a proof of concept. If you have more time later, please open a new pull request and we'll go from there. :slightly_smiling_face:

Issue #118, opened 2019-02-01
Refactor documentation into smaller pages (to make it more readable)


jwflory:
# Summary

Break up documentation from single, monolithic pages to more modular, bite-sized pages

# Background

We have documentation, but it is not easy to read. It is wordy and sometimes confusing. Our docs should take on small topics and focus on those topics. This also helps our documentation's searchability and makes it easier to navigate (in other words, important information isn't buried).

# Details

The installation page is in need of the most love. This page could be 3-4 pages. I was thinking to split it up this way:

* Quick installation guide (i.e. numbered steps about how to do a fast install)
* Create Telegram bot
* Configure IRC channel
* Explaining the config file
* Deployment options (currently `Usage` section)
* Maybe more…?

# Outcome

Documentation is more useful to non-developers, easier to find key information, better visibility into different parts of Teleirc

jwflory:
_Discussed in 2019-02-02 RITlug developers' meeting. This issue is targeted for the v1.3 milestone, estimated to release on March 2nd, 2019._

---

I volunteered to work on this during this sprint. It's something I've wanted to do for a long time. I'm targeting for completion the Saturday, Feb. 16th developers' meeting.

I'll probably start working on this in a feature branch gradually.

jwflory:
_Discussed in 2019-02-09 RITlug developers' meeting._

---

Not yet started. I plan to split off main installation page into several smaller pages. Expecting the main docs page to be split up by next developers' meeting on Saturday, Feb. 16th.

PR #121, opened 2019-02-15
docs: Add config file glossary

jwflory:
### Blocked by PR #120 – review it first!

---

Part of issue #118.

This commit adds a new page for a configuration file glossary. This page
is a highly-detailed overview of the config file and what each setting
does in the `env.example` file. It is intended for advanced users.

For now, this commit does not pull the duplicated parts out of
`installation.rst`, but I plan to deprecate that page in a future pull
request.

Preview of "Config file glossary" page

### DIFF
--- a/docs/config-file-glossary.rst
+++ b/docs/config-file-glossary.rst
@@ -1,0 +1,92 @@
+####################
+Config file glossary
+####################
+
+This page is a glossary of different settings in the ``env.example`` configuration file.
+The values shown for the settings are their defaults.
+This glossary is intended for advanced users.
+
+
+************
+IRC settings
+************
+
+``IRC_BLACKLIST=""``
+    Comma-separated list of IRC nicks to ignore
+
+``IRC_BOT_NAME=teleirc``
+    IRC nickname for bot
+
+``IRC_CHANNEL=#channel``
+    IRC channel for bot to join
+
+``IRC_SEND_STICKER_EMOJI=true``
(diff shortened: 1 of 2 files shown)


docs/config-file-glossary.rst
```diff
@@ -0,0 +1,92 @@
+####################
+Config file glossary
```
thenaterhood: My concern with config file documentation like this is that it decays over time because it gets forgotten. I understand the value of having it available in the documentation but that dimishes if it doesn't get updated with the options we support.

Would we be opposed to putting this information as comments in the config file itself?

xforever1313: ...Or somehow have this page be generated by something reading in the configuration file and its comments and outputting this page?

Tjzabel: I mean... *Technically* we could make a parser to read in the config file, and thus output this page. 

But... I think, since the `env.example` file doesn't get updated too often, we can keep the official documentation in the readthedocs. I'm not opposed to adding more comments into the `env.example` file in general, though. At least, maybe add a link to this readthedocs page instead of having `See the README for instructions.` at  the top of the file.

jwflory: @thenaterhood There might be a way to include the contents of a file. If we commented the example `env` file, we could use the file as living documentation. What do you think? We can open a new issue to track this.

jwflory:
I'm merging this, but I do want to track the conversation about using the file as living documentation in a new issue.

Issue #115, opened 2019-01-11
Image url reveals bot's token


alcir:
Imgur is not configured.
Sending an image on Telegram, the API URL appears on IRC.
Such URL contains the token of the bot, that should be keep private.

```'Untitled Image' uploaded by XXXXXX: https://api.telegram.org/file/botXXXXXXXXXXX:XXXXXXXXXXXXXXXXXX/photos/file_1.jpg```

Is this the only way to use the Telegram API?
There is a way to disable images on IRC, if Imgur is not configured?

Thanks.

xforever1313:
My guess is that files are also going to be impacted by this.

I think the best way to do this is add options to disable Photos and Files being sent via URL so we don't expose the bot API token.  These config items should probably be set to false by default, and have a comment with a warning about what happens if they are set to true (maybe even go so far as hiding them so its developers only).  I could see someone wanting to set this to true if they are testing the bot out, or if they are in a private IRC channel and trust everyone inside of it.

The good news is this should be straight forward to do since all of the TgHandlers take in an enabled boolean. We can just pass in the true/false from the config directly into the constructor.

jwflory:
@alcir @xforever1313 We suggested handling this issue in two ways.

## Option 1

1. Enabling Imgur support by default (with the generic API key provided by upstream library)
2. Supplementing docs to recommend getting your own API key to replace the default

This way, the Telegram API token is never exposed by accident.

## Option 2

1. Better documenting behavior of not using Telegram (i.e. API token will be exposed and this is a risk you are comfortable taking)

What do y'all think? @RITlug/teleirc-developers-commit-access

jwflory:
_Discussed in 2019-02-02 RITlug developers' meeting. This issue is targeted for the v1.3 milestone, estimated to release on March 2nd, 2019._

---

@Tjzabel is looking into this one. If @RITlug/teleirc-developers-commit-access have an opinion on previous options, please weigh in. :slightly_smiling_face: Otherwise, @Tjzabel is researching this more by the next developers' meeting on Saturday, Feb. 9th.

Tjzabel:
So as an update, I am looking to go with option 1.

This entails enabling Imgur by default, and writing documentation suggesting to get a personal Imgur API key. However, this does not fix the issue of when someone uploads an image as a *file*. When uploading as a file, the API key is still shown. 

In order to fix this, I think we can either:
1. Check to see if the photo is a file, and if so, don't make it go through to IRC side.
[...]

PR #127, opened 2019-02-22
TgDocuments Overhaul

Tjzabel:
The goal of this PR is to take care of:

1. Issue #115 
2. Enabling Imgur by default
3. Overhaul of TgDocumentHandler unit tests
4. Documentation surrounding these issues

### TL;DR
This PR removes the telegram bot's API from showing when images and files are sent over Telegram. Imgur is now enabled by default, and sending documents over Telegram is now set to `false` by default, and sends a message over to IRC with the document's caption so the IRC side isn't completely left out.

### DIFF
--- a/lib/TelegramHandlers/TgDocumentHandler.js
+++ b/lib/TelegramHandlers/TgDocumentHandler.js
@@ -30,14 +30,19 @@ class TgDocumentHandler {
      */
     async RelayDocumentMessage(from, document) {
         if (!this.Enabled) {
-            return;
-        }
+            let username = TgHelpers.ResolveUserName(from);
+            let message = username + 
+                        " shared a file on Telegram with caption: 'test'";
+            // TODO get caption of document 
 
-        let username = TgHelpers.ResolveUserName(from);
+            this._action(message);
 
-        let url = await this._tgbot.getFileLink(document.file_id);
+        } else {
+            let username = TgHelpers.ResolveUserName(from);
 
-        let message = username +
+            let url = await this._tgbot.getFileLink(document.file_id);
+
+            let message = username +
                         " Posted File: " +
                         document.file_name + 
                         " (" + 
@@ -47,7 +52,8 @@ class TgDocumentHandler {
                         " bytes): " +
                         url;
 
-        this._action(message);
+            this._action(message);
+        }
     }
 }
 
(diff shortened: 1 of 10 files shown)


jwflory:
Thanks @Tjzabel, this looks awesome. :100: I have some additional feedback:

1. **Code review**: Changes look good to me, but I'd prefer another @RITlug/teleirc-developers-commit-access developer to review too. I'm not as savvy.
2. **Before merging, a staging test**: To test these changes in a pre-existing install, I want to deploy this branch to our development bot in `#ritlug-teleirc`. (I'll approve the PR once I do this.)
3. **Consolidating commits**: Could you squash these commits to 2-3 commits with meaningful messages before we merge? Something like [these suggestions](https://juffalow.com/other/write-good-git-commit-message).

Thanks for working on this @Tjzabel. :+1:

Tjzabel:
@jwflory thanks for the review! I've squashed my commits into 2.

As a further note, here is what sending a document over to IRC looks like (disabled and enabled):

Disabled: 
```
Tjzabel21 shared a file on Telegram with caption: 'hello!'
```

Enabled: 
```
'My caption!' uploaded by Tjzabel21:
https://api.telegram.org/file/bot<api_key>/documents/<file>.jpg
```
* The `api_key` of your Telegram bot will be shown.

lib/TelegramHandlers/TgDocumentHandler.js
```diff
@@ -27,28 +28,41 @@ class TgDocumentHandler {
...
+     * @param {string} fileUrl - URL to the document
+    */
+    _GetDocumentMessage(caption, from, fileUrl) {
+        var message;
+
+        if (Helpers.IsNullOrUndefined(caption)) {
+            caption = "Untitled Document";
+        }
+
+        if (!this.Enabled) {
```
thenaterhood: I understand what you're doing here, but I want to confirm this is the behaviour we want since it's different from other handlers. From what I'm aware, other handlers don't send anything if they're disabled, whereas this one will send a notice that a document was sent without forwarding the document.

If we're cool with that, then so am I, it just seems inconsistent at first look and we should make sure it's documented.

Tjzabel: @thenaterhood yes, this is confirmed behavior, and is documented. 

Thanks for pointing that out!

jwflory:
We just tested this change on the staging environment in `#ritlug-teleirc`. Everything is looking good! Merging. :clapper:

PR #134, opened 2019-04-13
Bold usernames in message prefixes

### LINKED ISSUE #19: Add bold formatting to usernames in message strings on both Telegram and IRC
# Summary

When using the bridge bot, it would be helpful if usernames were marked up with bold text, to help separate them from the message body on IRC and Telegram.
# Analysis

In #8, there was a lot of discussion about how this should be implemented, especially in terms of using pre-existing libraries for markdown. Before completing or coming up with a solution to close this issue, please read through the comments and discussion in #8.

nic-hartley:
Resolves #19 

Note: There is no _officially_ standardized way to bold things in IRC. However, as far as I can tell, every major client supports wrapping things in `0x02` to make them bold, and the (few, but large) servers I've tested against all seem to handle it fine.

Telegram, on the other hand, does support bold higlighting, with "Markdown" parsing mode. It's not actually Markdown (it uses `*foo*` instead of `**foo**`, then `_foo_` to represent italics) but they call it that.


jwflory:
Thanks for reviewing @Tjzabel. :+1: Merging! :clapper:

PR #139, opened 2019-04-21
Complete QoL Improvements

Tjzabel:
## What this PR Does
* Adds ZWP into joining/parting Telegram messages
* Closes ongoing promise rejection handling
* Adds filetype IRC side when Telegram file sending set to false

### DIFF
--- a/lib/TelegramHandlers/TgHelpers.js
+++ b/lib/TelegramHandlers/TgHelpers.js
@@ -11,6 +11,7 @@ module.exports.ResolveUserName = function(from) {
     }
     
     // Add ZWP to remove IRC highlighting when Telegram + IRC nicks are the same
+    // See https://github.com/42wim/matterbridge/issues/175 for inspiration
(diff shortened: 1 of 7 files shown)


lib/TelegramHandlers/TgHelpers.js
```diff
@@ -25,5 +26,8 @@ module.exports.GetFullUserName = function(from) {
         return from.first_name;
     }
 
-    return from.first_name + " (@" + from.username + ")";
+    // Add ZWP so users with same username across bridge don't ping themselves
```
jwflory: Thanks for the useful comment, this is the kind of thing we would forget why we did it in two years.

package.json
```diff
@@ -4,6 +4,7 @@
   "description": "A simple Telegram <=> IRC bot",
   "main": "teleirc.js",
   "scripts": {
+    "start": "node teleirc.js",
```
jwflory: On a side note, it could be an interesting side-project for us to package an RPM for TeleIRC. A future sprint idea.

PR #141, opened 2019-04-21
Always write project name as TeleIRC (for branding)

jwflory:
This commit renames all uses of 'Teleirc' in our project to 'TeleIRC'.
This makes the name of our project match the cases of our new logo.
While this is a minor change, it better establishes the project identity
when we go to print materials for things like booth exhibits (e.g.
stickers, brochure hand-outs, etc.).

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -1,7 +1,7 @@
 #!/usr/bin/env python3
 # -*- coding: utf-8 -*-
 #
-# RITlug Teleirc documentation build configuration file, created by
+# RITlug TeleIRC documentation build configuration file, created by
 # sphinx-quickstart on Wed Mar 14 12:56:56 2018.
 #
 # This file is execfile()d with the current directory set to its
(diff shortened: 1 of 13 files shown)


docs/conf.py
```diff
@@ -152,9 +152,9 @@
 # (source start file, target name, title,
 #  author, documentclass [howto, manual, or own class]).
 latex_documents = [
-    (master_doc, 'RITlugTeleirc.tex', 'RITlug Teleirc Documentation',
+    (master_doc, 'RITlugTeleIRC.tex', 'RITlug TeleIRC Documentation',
```
thenaterhood: Seeing what appear to be filenames change with a PR like this makes me mildly nervous. I just want to double check that we intended and accounted for it :)

Tjzabel: @thenaterhood what makes you nervous?

jwflory: This appears to be setting the target file for the LaTeX to be written to. Since this is the output file name, I don't think it is an issue. Thanks for pointing it out either way.

PR #151, opened 2019-06-23
images: Pin Fedora 30 to Fedora Dockerfile

jwflory:
This commit does a couple of things:

* Pins Fedora Dockerfile to Fedora 30
* Switches to use `registry.fedoraproject.org` as the container source
* Splits packages installed across new lines (for readability)

I built this container locally and it worked. Ideally there needs to be
more work to make this viable (e.g. specifying config values as
environment variables), but that is out of scope for this pull request.

### DIFF
--- a/images/Dockerfile.fedora
+++ b/images/Dockerfile.fedora
@@ -1,4 +1,4 @@
-FROM fedora:29
+FROM registry.fedoraproject.org/fedora:30
 
 # Cache layers which will not change
 RUN groupadd teleirc -g 65532 \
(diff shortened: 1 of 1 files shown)


jwflory:
Oh, I also removed the `--setopt=deltarpm=false` argument as this was causing a strange segfault issue whenever I was building the container. :woman_shrugging: Since there is not a strong use case for us to require using this flag, I figured it was better to remove it.

images/Dockerfile.fedora
```diff
@@ -1,4 +1,4 @@
-FROM fedora:29
+FROM registry.fedoraproject.org/fedora:30
```
Tjzabel: ah, interesting! Nice to pull from the official fedora registry.

images/Dockerfile.fedora
```diff
@@ -7,8 +7,14 @@ RUN groupadd teleirc -g 65532 \
...
-RUN dnf -y upgrade --setopt=deltarpm=false \
-    && dnf -y install nodejs nodejs-yarn libicu-devel python gcc-c++ make --setopt=deltarpm=false \
+RUN dnf -y upgrade \
+    && dnf -y install \
+        nodejs \
+        nodejs-yarn \
+        libicu-devel \
+        python \
+        gcc-c++ \
+        make \
```
Tjzabel: Thanks for cleaning this up ^.^

PR #152, opened 2019-06-23
Reverse #134: Do not render IRC=>Telegram messages as Markdown

jwflory:
This commit is a partial reversal of the bold username feature
introduced in #134. This disables the Markdown rendering change for
messages sent from IRC to Telegram. Instead of only the username field
being rendered as Markdown, the entire message string is rendered. This
causes some messages to not send over the bridge if incomplete Markdown
is used and also mangles some URLs.

Eventually it would be nice to revisit this, but for now, this fixes a
critical bug at the expense of removing a feature on one half of the
bridge.

---

I also suggest a v1.3.2 bugfix release to follow this PR for the benefit of getting things working again for other users.

### DIFF
--- a/lib/IrcHandlers/IrcMessageHandler.js
+++ b/lib/IrcHandlers/IrcMessageHandler.js
@@ -56,7 +56,7 @@ class IrcMessageHandler {
         );
 
         if (blackListed === false) {
-            let message = "<*" + username + "*> " + userMessage;
+            let message = "<" + username + "> " + userMessage;
             this._action(message);
         }
     }
--- a/lib/TeleIrc.js
+++ b/lib/TeleIrc.js
@@ -332,7 +332,7 @@ class TeleIrc {
       this.config.tg.maxMessagesPerMinute,
       60,
       (message) => {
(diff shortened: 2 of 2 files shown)


jwflory:
Hmm. I ran the unit tests locally and they passed. I'll try to revisit why Travis is failing unrelated tests soon…

Tjzabel:
@jwflory tests are failing because the unit tests now check for the markdown additions. You'll have to remove that in order for tests to pass.

Tjzabel:
> Hmm. I ran the unit tests locally and they passed. I'll try to revisit why Travis is failing unrelated tests soon…

For what it's worth, I ran the tests locally off this branch as well, and it fails as expected.

Tjzabel:
@jwflory would it be better to simply do a full revert of the commit?

I'm having a hard time fixing all the unit tests otherwise.

Tjzabel:
I am actually going to close this issue as out of date. I would much prefer we work towards a solution instead of trying to do a revert, breaking our tests as well.

PR #155, opened 2019-07-19
Add nodemon as a development option.

### LINKED ISSUE #140: Consider supporting nodemon in development environments
## What this is
[Nodemon](https://nodemon.io/) is a nodejs tool to watch file changes, and restarts the service when any file changes occur. 

## What it looks like
`nodemon [--watch .env] teleirc.js`
By default, nodemon watches the current directory, excluding any dot files. You can add an optional `--watch` parameter to specify any other files (such as the `.env`) that are left out.

## Outcome
Using nodemon, if a change occurs to any file while TeleIRC is operating, the bot restarts, taking into account the newer changes. 

## How
* Install nodemon: `npm install nodemon`
* Replace `node` with `nodemon` in the startup

Tjzabel:
This PR adds nodemon as an option for development. Nodemon automatically restarts TeleIRC when configurations change, so developers don't need to keep stopping and restarting TeleIRC as they make changes. 

How to use:
```
nodemon teleirc.js
```

Furthermore, if you want to manually restart TeleIRC, you can simply type `rs` at any point, and it will quickly restart TeleIRC.

Closes #140.

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -61,6 +61,7 @@ See your IRC network's documentation on registering a channel.
 Change the `env.example` file to `.env`.
 Change the configuration values to the Telegram bot's tokens.
 For more help with configuration, see the [TeleIRC documentation](https://docs.teleirc.com/en/latest/quick-install/#configure-and-run-teleirc).
+Run TeleIRC with `npm start`, or optionally with [nodemon](https://nodemon.io/) to automatically restart TeleIRC when you make changes.
 
 
 ## Open a new pull request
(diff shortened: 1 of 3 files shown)


Tjzabel:
Hm, it seems nodemon may need some extra fandangling for node v8.

Tjzabel:
So I am going to close this issue. We would need to change how our CI works in order to get node v8 to build with nodemon, and I am fine with unofficially supporting this. It *does* make development a lot easier, but if I'm the only one who is going to be using it, and it breaks our node v8, I am okay with not supporting this feature.

PR #156, opened 2019-07-19
Move from Markdown to HTML formatting

### LINKED ISSUE #142: Escape IRC messages as plaintext (i.e. no Markdown formatting) in Telegram
# Summary

IRC messages sent to Telegram always render as plaintext (i.e. no Markdown formatting rules applied)

# Background

When hyperlinks including underscore characters are sent across the bridge, Telegram has suddenly started rendering this as italics:

`https://meetbot.fedoraproject.org/fedora-meeting-1/2019-04-25/f30-readiness-meeting.2019-04-25-19.00.log.html`

Rendered message in Telegram when URL is sent across TeleIRC

Obviously URLs should render as plain-text only.

# Details

This needs investigation. We need to do some message pre-processing on received IRC messages before sending them over to IRC. I think this is an unintended side effect of #134 when we changed `parse_mode`:

https://github.com/RITlug/teleirc/blob/b677658f32296e629e81867dfb814b4f4d308695/lib/TeleIrc.js#L335

# Outcome

URLs successfully appear as written when sent across bridge

Tjzabel:
## What This Does
This PR finally fixes #142, which covers issues we've had with IRC => Telegram markdown support, specifically with sending links that get transformed into markdown, rendering them unclickable Telegram-side. 

The parsing mode moves from markdown to HTML in order to take care of username bolding.

### Possible Issues
Doing HTML parsing may *possibly* cause issues, and so I think we should dig a little more into this before this is merged. However, all the functionality works just as intended, with only usernames being bolded, and links are sending just fine.

### Links
* [HTML Parsing](https://core.telegram.org/bots/api#html-style)
* [sendMessage object](https://core.telegram.org/bots/api#sendmessage)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,10 @@ node_js:
   - 8
   - 10
 
+cache:
+  npm: false
+  yarn: false
+
 addons:
   apt:
     update: true
@@ -20,7 +24,6 @@ script: yarn test
 notifications:
   irc:
     channels:
-      - "ircs://chat.freenode.net:6697/#rit-lug-sysadmin"
       - "ircs://chat.freenode.net:6697/#ritlug-teleirc"
     template:
       - "[%{repository_name}:%{branch}@%{commit} - build #%{build_number}] CI %{result}! (%{build_url})"
--- a/lib/IrcHandlers/IrcMessageHandler.js
+++ b/lib/IrcHandlers/IrcMessageHandler.js
@@ -1,4 +1,5 @@
 const Helpers = require("../Helpers.js");
+const escape = require('escape-html');
 
 /**
  * Handles the event when a user sends a message to the IRC channel.
@@ -56,7 +57,7 @@ class IrcMessageHandler {
         );
 
         if (blackListed === false) {
-            let message = "<*" + username + "*> " + userMessage;
+            let message = "&lt;<b>" + username + "</b>&gt; " + escape(userMessage);
             this._action(message);
         }
     }
--- a/lib/TeleIrc.js
+++ b/lib/TeleIrc.js
@@ -332,7 +332,7 @@ class TeleIrc {
       this.config.tg.maxMessagesPerMinute,
       60,
       (message) => {
-        teleirc.tgbot.sendMessage(teleirc.config.tg.chatId, message, { parse_mode: 'markdown' });
+        teleirc.tgbot.sendMessage(teleirc.config.tg.chatId, message, { parse_mode: 'html' });
       });
 
     this.tgbot.on('message', teleirc.tgEventListener.ParseMessage.bind(teleirc.tgEventListener));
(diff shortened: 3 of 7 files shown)


Tjzabel:
@jwflory thanks for reviewing! If/when this is merged, I would like to do a test run for a week(?) in `#ritlug-teleirc` *and* `#rit-foss` to get a better feel in a more active channel so we can catch errors a little quicker.

xforever1313:
My concern is what happens if a user sends a message with &lt; in it?  Or <b>hello</b>?  Will those be escaped?  If not, we should escape the user message.

But... JavaScript is a web technology, surely there must be a simple, standard library function to escape HTML.

*Googles*

...So JavaScript continues to let me down... we'll have to roll our own or pick a node module library to escape html >_>.

thenaterhood:
I detect a large quantity of salt in that comment, Seth. 🧂

Tjzabel:
@xforever1313 huh, funny thing with Riot, is that if you send `&lt;`, it automatically changes it to `<`. 

Sadly, the change to HTML is not feasible :frowning_face:. Any time *anyone* sends a `<` or some sort of HTML formatting string, TeleIRC will throw an error and the message will simply not send.

It looks like there is no way to effectively escape the formatting. @jwflory it looks like we may just have to completely disable any formatting from IRC => Telegram. Thoughts?

Tjzabel:
@xforever1313 this here looks simple enough: https://www.npmjs.com/package/escape-html

This may be the only addition we need around the [`userMessage`](https://github.com/RITlug/teleirc/blob/master/lib/IrcHandlers/IrcMessageHandler.js#L59) after including the library:
```
let message = "<*" + username + "*> " + escape(userMessage);
```

The only downside to the HTML formatting is we need to keep using `<` as our username prefix, because the formatting looks for lines starting with an HTML tag. This isn't an issue right now, since this is what we already do, but if we choose to move to a configurable prefix/suffix from IRC => Telegram like we do for Telegram => IRC, then we will need to look at a possible [escape-markdown](https://www.npmjs.com/package/escape-markdown) option instead.

Tjzabel:
Then again, the escaping will need further testing, because I'm not sure what would happen if someone sends an actual web link over.

Tjzabel:
@jwflory Heya! If you have time today, could you please take a look at my changes?

This PR successfully escapes all IRC => Telegram messages as plaintext :tada: :tada: 

The username bolding is done via HTML formatting, and the user's message itself is escaped, keeping the formatting intact. This means web links work again, any any odd formatting is completely removed, showing the user's message exactly as it was typed.

Tjzabel:
@thenaterhood just... *fine*? :frowning_face:

thenaterhood:
'spose I could have at least sprang for "good" haha

xforever1313:
👍

PR #171, opened 2019-09-15
Add new project layout, and TravisCI template

Tjzabel:
This PR is a mixture of mine and Nick Jones' commits. This PR:

* Sets up a basic Go project structure
* Contains a minimal TravisCI file

In the future, we may wish to consider finding a new CI platform (GitHub Actions?), as Travis was bought out and gutted several months ago. I'm not too sure what direction the company is headed in.


Tjzabel:
I took @Zedjones 's changes, and added a few tweaks to make sure the CI passes with a base `.travis.yml` file. 

Once we start adding basic unit tests, we should work towards creating a beefier and more sustainable CI system.

Tjzabel:
@jwflory Those are reasonable changes. I would actually like to look at deploying via Buildah/Podman down the road :smile:

Tjzabel:
@jwflory or should I do `deployments/images`? :thinking:

Zedjones:
So I'm gonna nitpick even more, but the items in the `deployments/images` directory aren't actually images, because images are the final product of running a `docker build` or equivalent. That's why I called the directory `docker`, since it contains instructions for docker/docker-compose on _how to create_ the images.

Tjzabel:
@jwflory would you be okay with moving the directory naming back to `deployments/containers`?

This method seems to be the most inclusive, as the `images` naming still falls under the same gripe, with images being what are built, and containers being what actually run.

PR #179, opened 2019-09-22
Resolves #174: Build out config parsing

### LINKED ISSUE #174: Build out config parsing
The first stage for creating a TeleIRC bot is reading in all the configuration settings, set as environment variables.

Zedjones:
This PR adds initial config parsing to the Go port, as well as removes an unneeded section from the `env.example` file that was related to a bug in a Node.js library. 

No testing was added at the moment due to indecision on how exactly we want to test the code.

### DIFF
--- a/cmd/teleirc.go
+++ b/cmd/teleirc.go
@@ -1,0 +1,16 @@
+package main
+
+import (
+	"fmt"
+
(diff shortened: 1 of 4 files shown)


Tjzabel:
Huh, so code climate thinks that this PR is merging into master? Not really sure how that's working

PR #181, opened 2019-09-22
Resolves #174: Build out config parsing

### LINKED ISSUE #174: Build out config parsing

Zedjones:
This PR adds initial config parsing to the Go port, as well as removes an unneeded section from the `env.example` file that was related to a bug in a Node.js library. 

No testing was added at the moment due to indecision on how exactly we want to test the code.

### DIFF
--- a/cmd/teleirc.go
+++ b/cmd/teleirc.go
@@ -1,0 +1,16 @@
+package main
+
+import (
+	"fmt"
+
+	"github.com/ritlug/teleirc/internal"
+)
+
+func main() {
+	settings, err := internal.LoadConfig("../env.example")
+	if err != nil {
+		fmt.Println(err)
+	} else {
+		fmt.Println(settings)
+	}
+}
--- a/env.example
+++ b/env.example
@@ -45,4 +45,4 @@ TELEGRAM_CHAT_ID=-0000000000000
 ###############################################################################
 
 USE_IMGUR_FOR_IMAGES=true
-IMGUR_CLIENT_ID=0000000000
+IMGUR_CLIENT_ID=7d6b00b87043f58
--- a/env.example
+++ b/env.example
@@ -46,14 +46,3 @@ TELEGRAM_CHAT_ID=-0000000000000
 
 USE_IMGUR_FOR_IMAGES=true
 IMGUR_CLIENT_ID=0000000000
-
-
-###############################################################################
-#                                                                             #
-#                      Misc. required env variables                           #
-#                -- not meant for user config, don't touch --                 #
-#                                                                             #
-###############################################################################
-
-# https://github.com/yagop/node-telegram-bot-api/issues/319#issuecomment-324963294
-NTBA_FIX_319=1
--- a/go.mod
+++ b/go.mod
@@ -1,3 +1,11 @@
 module github.com/Zedjones/teleirc
 
 go 1.12
+
+require (
+	github.com/caarlos0/env v3.5.0+incompatible
+	github.com/go-playground/locales v0.12.1 // indirect
+	github.com/go-playground/universal-translator v0.16.0 // indirect
+	github.com/go-playground/validator v9.29.1+incompatible
+	github.com/leodido/go-urn v1.1.0 // indirect
+)
--- a/go.mod
+++ b/go.mod
@@ -1,0 +1,3 @@
+module github.com/Zedjones/teleirc
+
+go 1.12
--- a/go.mod
+++ b/go.mod
@@ -1,4 +1,4 @@
-module github.com/Zedjones/teleirc
+module github.com/ritlug/teleirc
 
 go 1.12
 
--- a/internal/config.go
+++ b/internal/config.go
@@ -1,5 +1,12 @@
 package internal
 
+import (
+	"github.com/caarlos0/env"
+	"github.com/go-playground/validator"
+)
+
+var validate *validator.Validate
+
 // IRCSettings includes settings related to the IRC bot/message relaying
 type IRCSettings struct {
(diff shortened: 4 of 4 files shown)


go.mod
```diff
@@ -0,0 +1,11 @@
+module github.com/Zedjones/teleirc
```
Tjzabel: This should be changed :)

internal/config.go
```diff
@@ -0,0 +1,86 @@
...
+	BotName             string   `env:"IRC_BOT_NAME" envDefault:"teleirc"`
+	SendStickerEmoji    bool     `env:"IRC_SEND_STICKER_EMOJI" envDefault:"true"`
+	SendDocument        bool     `env:"IRC_SEND_DOCUMENT" envDefault:"true"`
+	Prefix              string   `env:"IRC_PREFIX" envDefault:"<"`
+	Suffix              string   `env:"IRC_SUFFIX" envDefault:">"`
+	ShowJoinMessage     bool     `env:"IRC_SHOW_JOIN_MESSAGE" envDefault:"true"`
+	ShowLeaveMessage    bool     `env:"IRC_SHOW_LEAVE_MESSAGE" envDefault:"true"`
+	NickServPassword    string   `env:"IRC_NICKSERV_PASS" envDefault:""`
+	NickServService     string   `env:"IRC_NICKSERV_SERVICE" envDefault:""`
+	EditedPrefix        string   `env:"IRC_EDITED_PREFIX" envDefault:"[EDIT]"`
```
Tjzabel: Super minor, but `"[EDIT] "` requires a space after the closing bracket

internal/config.go
```diff
@@ -0,0 +1,86 @@
...
+	ShowActionMessage   bool   `env:"SHOW_ACTION_MESSAGE" envDefault:"false"`
+	ShowLeaveMessage    bool   `env:"SHOW_LEAVE_MESSAGE" envDefault:"false"`
+	ShowKickMessage     bool   `env:"SHOW_KICK_MESSAGE" envDefault:"false"`
+	MaxMessagePerMinute int    `env:"MAX_MESSAGE_PER_MINUTE" envDefault:"20"`
+}
+
+// ImgurSettings includes settings related to Imgur uploading for Telegram photos
+type ImgurSettings struct {
+	UseImgurForImageLinks bool   `env:"USE_IMGUR_FOR_IMAGE" envDefault:"true"`
+	ImgurClientID         string `env:"IMGUR_CLIENT_ID" envDefault:"12345"`
```
Tjzabel: default will have to change for the Client ID. Please replace this with `7d6b00b87043f58`. This will be our default TeleIRC Imgur token.

cmd/teleirc.go
```diff
@@ -0,0 +1,16 @@
...
+)
+
+func main() {
+	settings, err := internal.LoadConfig("../env.example")
+	if err != nil {
+		fmt.Println(err)
+	} else {
+		fmt.Println(settings)
+	}
+}
```
Tjzabel: This is essentially a dummy file, so I'm okay with merging this in for now

env.example
```diff
@@ -45,15 +45,4 @@ TELEGRAM_CHAT_ID=-0000000000000
...
-###############################################################################
-#                                                                             #
-#                      Misc. required env variables                           #
-#                -- not meant for user config, don't touch --                 #
-#                                                                             #
-###############################################################################
-
-# https://github.com/yagop/node-telegram-bot-api/issues/319#issuecomment-324963294
-NTBA_FIX_319=1
+IMGUR_CLIENT_ID=0000000000
```
Tjzabel: I left a note elsewhere, but please change this default to `7d6b00b87043f58`, as this is what we are going to use as TeleIRC's default Imgur token.

go.mod
Zedjones: Whoopsie, good catch. That shouldn't be in there at all haha

internal/config.go
Zedjones: Appreciated, thanks

cmd/teleirc.go
```diff
@@ -0,0 +1,16 @@
+package main
+
+import (
+	"fmt"
+
+	"github.com/ritlug/teleirc/internal"
+)
+
+func main() {
+	settings, err := internal.LoadConfig("../env.example")
```
jwflory: This is late, but if I read this right, is it loading the `env.example` file itself as the config file? What if the user has a non-default config, e.g. `.env`? Maybe I'm not reading this right?

env.example
```diff
@@ -45,15 +45,4 @@ TELEGRAM_CHAT_ID=-0000000000000
...
-###############################################################################
-#                                                                             #
-#                      Misc. required env variables                           #
-#                -- not meant for user config, don't touch --                 #
-#                                                                             #
-###############################################################################
-
-# https://github.com/yagop/node-telegram-bot-api/issues/319#issuecomment-324963294
-NTBA_FIX_319=1
+IMGUR_CLIENT_ID=7d6b00b87043f58
```
jwflory: Just curious, whose Imgur client ID is this? @Zedjones I assume?

Tjzabel: @jwflory this is the TeleIRC default ID.

cmd/teleirc.go
Tjzabel: This isn't in the code anymore :upside_down_face:

env.example
jwflory: Right. I mean to whose account the API key is registered to. It would be good to take note.

Tjzabel: This API key is registered to the RITlug Imgur account. Password information is stored in RITlug's private GItLab.

PR #187, opened 2019-09-26
Add config glossary 

Tjzabel:
This PR updates some outstanding fields available in the `env.example` that are not present in our config glossary documentation.


jwflory:
I appreciate the documentation love. :heart:

PR #188, opened 2019-09-26
Add base logic connecting IRC and Telegram

Tjzabel:
## What This PR Does
* Builds out main `teleirc.go`
* Adds basic framework for IRC and Telegram handlers
* Adds connection to Config Settings


cmd/teleirc.go
```diff
@@ -1,16 +1,45 @@
...
+	}
+
+	fmt.Println(tgbot)
+}
+
 func main() {
-	settings, err := internal.LoadConfig("../env.example")
+	// optional path flag if user does not want .env file in root dir
+	var path string
+	flag.StringVar(&path, "p", "../.env", "Path to .env")
```
Zedjones: I think setting the default value to ".env" might make more sense, given that we're usually going to specify our own env file during development.

Tjzabel: yeah that seems to be my general thought as well.

Tjzabel:
@Zedjones I updated the PR to reflect your issues.

Tjzabel:
@robbyoconnor thanks for reviewing :+1:

PR #197, opened 2019-10-01
Resolves #189: Initial implementation of IRC client

### LINKED ISSUE #189: Create initial IRC bot using Config Settings

Zedjones:


internal/handlers/irc/irc.go
```diff
@@ -1,9 +1,69 @@
...
+}
+
 /*
-StartBot instantiates an IRC bot.
+connectHandler return a function to use as the connect handler for girc,
+so that the specified channel is joined after the server connection is established
 */
-func StartBot() (string, error) {
-	return "Hello from IRC!", nil
+func connectHandler(c Client) func(*girc.Client, girc.Event) {
```
robbyoconnor: Should this be `connectionHandler`

Zedjones: I think `connectHandler` makes more sense, as it's the handler that gets fired when the bot connects to the server.

Zedjones:
> Should we use [viper](https://github.com/spf13/viper) for our configuration file handling vs rolling our own?

We're not rolling our own, we're using a combination of `env`, `godotenv`, and `validator`. You can look at `config.go` to see how it's being done. We considered Viper but it's overly heavy for what we're doing and doesn't support any sort of struct tag based population of fields or validation.

internal/handlers/irc/irc.go
```diff
@@ -1,9 +1,69 @@
...
+}
+
+/*
+addHandlers adds handlers for the client struct based on the settings
+that were passed in to NewClient
+*/
+func (c Client) addHandlers() {
+	c.Handlers.Add(girc.ALL_EVENTS, func(c *girc.Client, e girc.Event) {
+		fmt.Println(e.String())
+	})
```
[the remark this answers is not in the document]
Zedjones: Yup, that is correct. So when the server pings the client, when a message is sent, somebody joins the channel, etc.

PR #200, opened 2019-10-06
Resolves #199: Add concurrency to the IRC bot

### LINKED ISSUE #199: Build concurrency into the initialization of the IRC bot
As it stands right now, after calling `irc.StartBot()`, nothing will run afterwards, as it is an ongoing function call. We need to fix this so it utilizes something like channels, so we can continue onto the next function call while it is running.

Zedjones:


cmd/teleirc.go
```diff
@@ -27,13 +27,20 @@ func main() {
...
+	}
+
+	startTelegram()
+	client := irc.NewClient(settings.IRC)
+
+	ircChan := make(chan error)
+	go client.StartBot(ircChan)
+
+	select {
+	case err := <-ircChan:
```
Tjzabel: Are cases the best way to handle this? Are there any docs on that?

Zedjones: Sure, [here's](https://gobyexample.com/non-blocking-channel-operations) an example from gobyexample.com. Pretty standard, allows you to block until one of the two channels returns and then executes the respective case statement.

Zedjones: In this case, of course, there's only one until the Telegram bot has a channel for sending messages as well

Zedjones: This use case is what `select` is made for, it's not the same as a switch statement

Zedjones:
Just to give more context around what I'm doing, this is what I changed:

1. `go client.StartBot(ircChan)` starts the bot in a separate "thread", more accurately a separate coroutine.
2. The channel passed in allows the `StartBot` function to send messages to the parent coroutine in `teleirc.go`.
3. All that I really had to do for this was edit the returns in `StartBot` to send messages across the channel instead (the parameter `chan<- error` means that the argument can only be used to send errors, not receive them.
4. The `select` statement takes a number of `case` statements that include channel operations. In our case, we want to wait until a message is sent over the ircChan (since this means the coroutine exited and returned either nil or an error) or tgChan (doesn't exist yet).

Here are some resources:
- [Coroutines](https://gobyexample.com/goroutines)
- [Channels](https://gobyexample.com/channels)
- [Select](https://gobyexample.com/select)

Tjzabel:
I apologize for the long delay in reviewing this :frowning_face: 

I *should* be able to take a look in the next few days.

Tjzabel:
@jwflory thanks for merging this through :+1:

cmd/teleirc.go
Tjzabel: @Zedjones thanks for the example. I'll take a look at this info when I can :+1:

PR #203, opened 2019-10-18
Build out concurrent Telegram bridge

Tjzabel:
# What this PR Does
1. Brings concurrency to Telegram :tada: :tada: 
2. Adds additional flag support


Tjzabel:
I've since added output when setting up TeleIRC so the user knows what processes are happening:
```
Creating new Telegram bot client...
Creating new IRC bot client...
Starting up Telegram bot...
Startup up IRC bot...
Adding IRC event handlers...
```

I would like to build this out (not in this PR) further to let the user know when TeleIRC is fully set up and runniing. This seems like we would need to have a way to tie all the startup goroutines together so we know when they've all completed their tasks. I'm not sure yet what the best implementation would be, but I think it would be very useful to look into.

env.example
```diff
@@ -8,7 +8,7 @@
 
 IRC_BLACKLIST=""
 IRC_BOT_NAME=teleirc
-IRC_CHANNEL=#channel
+IRC_CHANNEL="#channel"
```
Tjzabel: @jwflory @Zedjones quotations needed to be added here in the `env.example` because the config parsing sees the `#` as a comment, therefore ignoring the IRC channel's name.

Tjzabel:
pl0x review

env.example
robbyoconnor: LGTM

Tjzabel:
@robbyoconnor woops, thanks for the typo fix!

As for this PR, it's now been two weeks since this PR went out. Since I need this merged through for all the other issues that require this functionality, I hope you all don't mind if I merge my own PR through. 

The next steps after this PR is merged are:
1. Rearrange TeleIRC go port to only include explicit dependencies
    * This seems to be the best way for Go unit tests to work, more on this at our developer meeting
2. Write (or rewrite) unit tests for the initial Telegram startup and config testing.
3. Determine the best method for crossing over the two platforms.
4. Determine the Handler hierarchy.


Furthermore, I would like to use the Logger package for the initial startup logging, and remove the `fmt.Println` statements. I am looking to add this as we complete number one on the list above.

cmd/teleirc.go
```diff
@@ -10,37 +10,46 @@ import (
...
 
-	settings, err := internal.LoadConfig(path)
+	if *flagVersion {
+		fmt.Printf("Current TeleIRC version: %s\n", version)
+		return
+	}
+
+	// TODO: Build out debugging capabilities for more verbose output
+	if *flagDebug {
+		fmt.Printf("Debug mode currently set to: %s\n", flagDebug)
```
Zedjones: `go-vet` complains about this being `%s` instead of `%t`, and I believe `flagDebug` should be `*flagDebug`.

cmd/teleirc.go
```diff
@@ -10,37 +10,46 @@ import (
...
 		return
 	}
 
-	startTelegram()
-	client := irc.NewClient(settings.IRC)
+	tgClient, _ := tg.NewClient(settings.Telegram)
+	tgChan := make(chan error)
+	go tgClient.StartBot(tgChan)
 
+	IrcClient := irc.NewClient(settings.IRC)
```
Zedjones: Small thing, but I think keeps this consistent with camelCase would make it `ircClient` instead

Tjzabel:
@Zedjones I made the changes you requested. If this looks good, can you please approve the PR?

Zedjones:
If an invalid API key is specified (for example, the default value), this happens:
```
Creating new Telegram bot client...
Error creating Telegram client
Creating new IRC bot client...
Starting up Telegram bot...
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x562758fe512c]

goroutine 29 [running]:
github.com/go-telegram-bot-api/telegram-bot-api.(*BotAPI).GetUpdatesChan(0x0, 0x0, 0x0, 0x3c, 0x1, 0x1c, 0x0)
        /home/zedjones/go/pkg/mod/github.com/go-telegram-bot-api/telegram-bot-api@v4.6.4+incompatible/bot.go:484 +0x5c
github.com/ritlug/teleirc/internal/handlers/telegram.(*Client).StartBot(0xc000286880, 0xc0002da000)
        /home/zedjones/Documents/RITLug/tjzabel/teleirc/internal/handlers/telegram/telegram.go:42 +0xe7
created by main.main
        /home/zedjones/Documents/RITLug/tjzabel/teleirc/cmd/teleirc.go:43 +0x30c
exit status 2
```
I'm not sure how exactly we can do this better, but we probably don't want an end user to see this.

PR #204, opened 2019-10-25
Remove asserts from unit tests

Tjzabel:
This PR removes assert statements from unit tests. This is more of a discussion point for figuring out how we wish to proceed with unit testing. Since I would like to follow as closely as we can to the Go standard library, I would like to personally follow this standard of unit testing development. 

Furthermore, this PR removes the `LoadConfig` logic surrounding the loading of the environment file based on the path. I would like `LoadConfig` to make the assumption that the path exists, since the default value of `path` will always be `.env`. Also, I found the logic unnecessary because `godotenv.Load` already handles cases when the path is not found in a pretty straightforward way.


Tjzabel:
Hmm, so I really don't care which way we go with using asserts or not. 

However, I do want the changes to the `config.go` to be reviewed, since I removed a bunch of error checking. This was done due to `godotenv.Load()` handling these errors for us, so checking if the path exists isn't something that we need to do on our end.

Tjzabel:
@Zedjones @jwflory I've simplified this PR to only contain the `LoadConfig` changes. I am perfectly fine with continuing to use asserts in our unit testing.

Zedjones:
Would it be possible to add the flag for using environment variables instead of the file? Or perhaps make an issue so we don't forget about this?

Tjzabel:
@Zedjones hmm, it might make more sense to make an issue for it, since we would have to change more than just adding a flag for it

Tjzabel:
oh. I uh, guess I deleted the wrong branch on my fork :sweat_smile: 

My own fork was pretty out of date, after realizing I haven't been pushing changes locally into my fork. Instead of reopening this PR, I'll make an issue to discuss this further.

PR #217, opened 2019-11-04
Improve documentation around contributing and workflow

jwflory:
This PR is opened to address feedback and discussion in #194. It includes a collection of improvements and reorganizations of information in the new docs format. To summarize, here are the commit messages:

* **docs: Move contributing guidelines into docs (closes #194)** — 1e64881
* **docs: Move FAQ/who uses pages into their own category** — 227fa05
* **docs: Add recommonmark as docs dependency** — e1be5ab
* **docs: Add placeholder file to _static** — 1a1d4f5
* **docs: Add placeholder CONTRIBUTING.md file** — cdf11a5

PR #218, opened 2019-11-04
ci: Add job to build and test docs

jwflory:
This commit switches our Travis CI builds over to a matrix build system.
In addition to performing the standard unit test suite for NodeJS, it
will also kick up a VM to build and test our docs. This is a fairly
lightweight way to guarantee that our docs are building.

Also, you will notice this commit is failing! That is expected. The
`recommonmark` dependency was not added in #215, but it was added in
PR #217. Once that PR merges, this CI test will pass again.


jwflory:
I rebased after #217 merged, and ta-da, CI passes! :sunglasses: Merging! :ocean:

PR #220, opened 2019-11-06
Telegram unit tests and explicit dependency ordering

### LINKED ISSUE #191: Add unit tests to validate main TeleIRC bot
At this point, we need to start creating unit tests for config parsing (if needed), and the main TeleIRC file.

Tjzabel:
# PR Overview
This PR hopes to address the explicit dependencies issue from the Telegram side, and adds unit tests to verify a new Telegram client is properly created with the given config settings.

## PR Specifics
* Moves client connection into `StartBot()`
* Sets Telegram client API key from the Client struct, and not from the outside library itself
* Creates unit tests to make sure the client is properly set up with the correct, given settings

## Other Notes
With these unit tests done, I don't believe there is a reason to implement unit tests directly on the `cmd/teleirc.go`. So, with the IRC and Telegram base unit tests done, this PR closes #191.

### DIFF
--- a/cmd/teleirc.go
+++ b/cmd/teleirc.go
@@ -38,12 +38,7 @@ func main() {
 		return
 	}
 
-	tgClient, err := tg.NewClient(settings.Telegram)
-	if err != nil {
-		fmt.Println(err)
-		return
-	}
-
+	tgClient := tg.NewClient(settings.Telegram)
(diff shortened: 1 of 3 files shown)


Zedjones:
Everything looks good, unit tests passed and both bots properly echo messages still :+1:

PR #222, opened 2019-11-16
Fix config parsing

felixonmars:
The `|| true` lines make corresponding options to be always true, one cannot falsify the parser by setting it to "false" in `.env`. The other `|| false` lines are just redundant.


Tjzabel:
I'm merging this through. If you'd like to use this feature before our next release, feel free to base off the master branch :+1:

jwflory:
@felixonmars Thanks for contributing! :+1:

PR #225, opened 2019-12-27
Fix link to "Who uses TeleIRC"

JoshuaMulliken:
The link as it is currently is broken. I have changed it to the correct link.


jwflory:
@JoshuaMulliken Thanks for catching and fixing this! :+1:

PR #230, opened 2020-01-09
Implement handler "interface" and add function for mapping handlers to IRC

Zedjones:
This adds a type alias for IRC handlers in a `handlers.go` file within `internal/handlers/irc` and adds a function within that returns a map of strings (girc event types) to handler functions. This will be edited as handlers are created, and then the handlers get added to the client as such
```Go
for eventType, handler := range c.HandlerMap {
    c.Handlers.Add(eventType, handler(c))
}
```


cmd/teleirc.go
```diff
@@ -4,16 +4,19 @@ package main
...
+	tgbotapi "github.com/go-telegram-bot-api/telegram-bot-api"
 	"github.com/ritlug/teleirc/internal"
 	"github.com/ritlug/teleirc/internal/handlers/irc"
 	tg "github.com/ritlug/teleirc/internal/handlers/telegram"
 )
 
-var (
+const (
 	version = "v2.0"
+)
```
jwflory: Good call :+1:

cmd/teleirc.go
```diff
@@ -34,8 +37,8 @@ func main() {
 
 	settings, err := internal.LoadConfig(*flagPath)
 	if err != nil {
-		fmt.Println(err)
-		return
+		fmt.Fprintln(os.Stderr, err)
+		os.Exit(1)
```
jwflory: Aha – is this the exit code we talked about, in the context of systemd service files?

Zedjones: Yep! Eventually we'll change it to use a proper logger, but this is better than nothing for the moment.

internal/handlers/irc/irc.go
```diff
@@ -64,20 +70,7 @@ that were passed in to NewClient
...
-}
-
-/*
-connectHandler return a function to use as the connect handler for girc,
-so that the specified channel is joined after the server connection is established
-*/
-func connectHandler(c Client) func(*girc.Client, girc.Event) {
-	return func(gc *girc.Client, e girc.Event) {
-		c.Cmd.Join(c.Settings.Channel)
+	for eventType, handler := range getHandlerMapping() {
```
Tjzabel: nice way of doing this :smile:

Tjzabel: Nice way of doing this :smile:

internal/handlers/irc/handlers.go
```diff
@@ -0,0 +1,43 @@
...
+			c.SendMessage(formatted)
+		}
+	}
+}
+
+/*
+getHandlerMapping returns a mapping of girc event types to handlers
+*/
+func getHandlerMapping() map[string]Handler {
+	return map[string]Handler{
```
Tjzabel: ohhh, fancy

internal/handlers/irc/handlers.go
```diff
@@ -0,0 +1,43 @@
...
+
+/*
+Handler specifies a function that handles an IRC event
+In this case, we take an IRC client and return a function that
+handles an IRC event
+*/
+type Handler = func(c Client) func(*girc.Client, girc.Event)
+
+/*
+connectHandler return a function to use as the connect handler for girc,
```
Tjzabel: super pedantic, but "connectHandler return**s** a function..."

.gitignore
```diff
@@ -30,3 +30,7 @@ pids
 
 # Running config file! Don't commit this unless you know what you're doing.
 .env
+
+# Generated executables
+teleirc
+cmd/cmd
```
Tjzabel: For some reason VS Code doesn't put EOF/newlines at the end of the file, so GitHub adds a symbol to let us know. Not sure if this hurts anything?

internal/handlers/irc/irc.go
Zedjones: Thank you! 😄

.gitignore
Zedjones: It shouldn't hurt anything, it's just a stylistic preference I think? Maybe it makes a difference on Windows? Not sure

Zedjones:
Gonna go ahead and merge this, thanks for the reviews :+1:

PR #232, opened 2020-01-12
Implemented IRC joining/parting handlers

Zedjones:
Implemented handlers for joining and parting, only thing I'm unsure of is if [QUIT](https://tools.ietf.org/html/rfc2812#section-3.1.7) qualified as parting or if we treat that separately.

### DIFF
--- a/internal/handlers/irc/handlers.go
+++ b/internal/handlers/irc/handlers.go
@@ -9,6 +9,7 @@ import (
 const (
 	joinFmt = "* %s joins"
 	partFmt = "* %s parts"
+	quitFmt = "* %s quit (%s)"
 )
 
 /*
@@ -28,9 +29,13 @@ func connectHandler(c Client) func(*girc.Client, girc.Event) {
 	}
 }
 
-func privMsgHandler(c Client) func(*girc.Client, girc.Event) {
+/*
+messageHandler handles the PRIVMSG IRC event, which entails both private
+and channel messages. However, it only cares about channel messages
+*/
+func messageHandler(c Client) func(*girc.Client, girc.Event) {
 	return func(gc *girc.Client, e girc.Event) {
-		formatted := fmt.Sprintf(c.MessageFormat, e.Source.Name, e.Params[1])
+		formatted := c.Settings.Prefix + e.Source.Name + c.Settings.Suffix + " " + e.Params[1]
 		if e.IsFromChannel() {
 			c.SendMessage(formatted)
 		}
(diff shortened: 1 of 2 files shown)


internal/handlers/irc/handlers.go
```diff
@@ -6,6 +6,11 @@ import (
 	"github.com/lrstanley/girc"
 )
 
+const (
+	joinFmt = "<system> %s joins"
+	partFmt = "<system> %s parts"
```
jwflory: It's a stylistic preference, but I assume this is sent in Telegram when an IRC user parts/joins:

`<system> jwflory joins`

My preference would be something like:

`* jwflory joins`

It's my preference, but I won't make this suggestion blocking.

Tjzabel:
As an update, @Zedjones has agreed to make the `QUIT` functionality before putting this PR through.

Zedjones:
Also need to change format for message string to be more readable

internal/handlers/irc/handlers.go
```diff
@@ -23,21 +29,46 @@ func connectHandler(c Client) func(*girc.Client, girc.Event) {
 	}
 }
 
-func privMsgHandler(c Client) func(*girc.Client, girc.Event) {
+/*
+messageHandler handles the PRIVMSG IRC event, which entails both private
+and channel messages. However, it only cares about channel messages
+*/
+func messageHandler(c Client) func(*girc.Client, girc.Event) {
```
Tjzabel: Neat, thanks for renaming and explaining

internal/handlers/irc/handlers.go
```diff
@@ -23,21 +29,46 @@ func connectHandler(c Client) func(*girc.Client, girc.Event) {
 	}
 }
 
-func privMsgHandler(c Client) func(*girc.Client, girc.Event) {
+/*
+messageHandler handles the PRIVMSG IRC event, which entails both private
+and channel messages. However, it only cares about channel messages
```
Tjzabel: I'm guessing this means we only handle channel messages?

internal/handlers/irc/handlers.go
```diff
@@ -23,21 +29,46 @@ func connectHandler(c Client) func(*girc.Client, girc.Event) {
...
+*/
+func messageHandler(c Client) func(*girc.Client, girc.Event) {
 	return func(gc *girc.Client, e girc.Event) {
-		formatted := fmt.Sprintf(c.MessageFormat, e.Source.Name, e.Params[1])
+		formatted := c.Settings.Prefix + e.Source.Name + c.Settings.Suffix + " " + e.Params[1]
```
Tjzabel: so this is fine, but we've actually historically made the `SUFFIX` variable set to `"> "` so we don't need to worry about the extra `+ " "` you have in there.

internal/handlers/irc/handlers.go
```diff
@@ -23,21 +29,46 @@ func connectHandler(c Client) func(*girc.Client, girc.Event) {
...
 /*
 getHandlerMapping returns a mapping of girc event types to handlers
 */
 func getHandlerMapping() map[string]Handler {
 	return map[string]Handler{
 		girc.CONNECTED: connectHandler,
-		girc.PRIVMSG:   privMsgHandler,
+		girc.PRIVMSG:   messageHandler,
+		girc.PART:      partHandler,
+		girc.QUIT:      quitHandler,
```
Tjzabel: it would be *so great* if this was alphabetized

Issue #212, opened 2019-11-03
Implement Unified Message Handling between IRC and Telegram


Tjzabel:
Here is the outcome of our discussions on how to best create the unfiied structure:
```Go
(...)

tgClient, _ := tg.NewClient(settings.Telegram)
tgChan := make(chan error)

ircClient := irc.NewClient(settings.IRC)
ircChan := make(chan error)

go ircClient.StartBot(ircChan, tg.SendMessage)
go tgClient.StartBot(tgChan, irc.SendMessage)

(...)
```

_Originally posted by @Tjzabel in https://github.com/RITlug/teleirc/issues/198#issuecomment-549101020_

jwflory:
@Tjzabel Are you targeting this issue for the current sprint or a next sprint? Please update this card on the project board when you have a chance. Thanks!

jwflory:
No new updates over break. This issue is a focus of the next sprint.

The Telegram side is somewhat more complicated because of third-party library. @Tjzabel is researching how sending is done on Telegram to higher level of abstraction, so we can send any text object across the bridge.

## Closing criteria

- [ ] Send simple text message from one platform to the other

---

Related to #210.

Tjzabel:
We now have a complete TeleIRC bridge working in Go :tada: :tada:

PR #235, opened 2020-01-24
Allow communication from Telegram => IRC 

Tjzabel:
## What This PR Does
Relays messages from Telegram => IRC.
Part 1/2 of issue #212 

## Next Steps
@Zedjones is tasked with creating the IRC => Telegram relay.

### DIFF
--- a/internal/config.go
+++ b/internal/config.go
@@ -37,6 +37,8 @@ type IRCSettings struct {
 type TelegramSettings struct {
 	Token               string `env:"TELEIRC_TOKEN,required"`
 	ChatID              int64  `env:"TELEGRAM_CHAT_ID,required"`
+	Prefix              string `env:"TELEGRAM_MESSAGE_PREFIX" envDefault:"<"`
+	Suffix              string `env:"TELEGRAM_MESSAGE_SUFFIX" envDefault:"> "`
 	ShowJoinMessage     bool   `env:"SHOW_JOIN_MESSAGE" envDefault:"false"`
 	ShowActionMessage   bool   `env:"SHOW_ACTION_MESSAGE" envDefault:"false"`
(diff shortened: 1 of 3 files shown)


internal/handlers/telegram/telegram.go
```diff
@@ -60,14 +62,12 @@ func (tg *Client) StartBot(errChan chan<- error) {
 		if update.Message == nil {
 			continue
 		}
-		//formatted := "<" + update.Message.From + "> " + update.Message.Text
-		//fmt.Println("Sending message:", update.Message.Text)
-		//tg.SendMessage(formatted)
+		formatted := tg.Settings.Prefix +
+			update.Message.From.UserName +
+			tg.Settings.Suffix +
+			update.Message.Text
```
Zedjones: Might be cleaner to use a format string here, especially since we already have the prefix and suffix at client creation time. Not going to block on this, but it's worth cleaning up imo

internal/config.go
```diff
@@ -37,6 +37,8 @@ type IRCSettings struct {
 type TelegramSettings struct {
 	Token               string `env:"TELEIRC_TOKEN,required"`
 	ChatID              int64  `env:"TELEGRAM_CHAT_ID,required"`
+	Prefix              string `env:"TELEGRAM_MESSAGE_PREFIX" envDefault:"<"`
+	Suffix              string `env:"TELEGRAM_MESSAGE_SUFFIX" envDefault:"> "`
```
Zedjones: So are these two settings going to be separate from the IRC settings now?

Tjzabel: yeah, these settings are what shows up on IRC from Telegram.

Tjzabel:
@Zedjones can you review this again? I don't think you ever explicitly approved it

PR #254, opened 2020-03-01
fmt print statements replaced using the log package

### LINKED ISSUE #245: Migrate existing print statements to use Go logging library
# Summary

Migrate existing print statements to use Go's standard library for logging

# Background

Currently the only logging we use in the Go port are print statements.

Per #213, we should use the Go standard library for logging. This will give us more flexibility in verbosity levels and formatting later on.

We considered `logrus` but decided to keep it simple and use the standard library to keep our dependencies down.

# Details

This is where I could find any existing use of print statements in our code:

```bash
$ grep --exclude-dir={docs/,.git/} -rnw . -i -e '.*print.*'

./cmd/teleirc.go:29:		fmt.Printf("Current TeleIRC version: %s\n", version)
./cmd/teleirc.go:35:		fmt.Printf("Debug mode currently set to: %t\n", *flagDebug)
./cmd/teleirc.go:40:		fmt.Fprintln(os.Stderr, err)
./cmd/teleirc.go:55:		fmt.Println(ircErr)
./cmd/teleirc.go:57:		fmt.Println(tgErr)
./internal/handlers/telegram/telegram.go:25:	fmt.Println("Creating new Telegram bot client...")
./internal/handlers/telegram/telegram.go:43:	fmt.Println("Starting up Telegram bot...")
./internal/handlers/telegram/telegram.go:47:		fmt.Println("Failed to connect to Telegram")
./internal/handlers/irc/handlers.go:47:		c.sendToTg(fmt.Sprintf(joinFmt, e.Source.Name))
./internal/handlers/irc/handlers.go:53:		c.sendToTg(fmt.Sprintf(partFmt, e.Source.Name))
./internal/handlers/irc/handlers.go:59:		c.sendToTg(fmt.Sprintf(quitFmt, e.Source.Name, e.Params[0]))
./internal/handlers/irc/irc.go:24:	fmt.Println("Creating new IRC bot client...")
./internal/handlers/irc/irc.go:45:	fmt.Println("Starting up IRC bot...")
./internal/handlers/irc/irc.go:68:	fmt.Println("Adding IRC event handlers...")
```

# Outcome

Consistent use of the Go library for logging where we are currently using print statements

10eMyrT:
Closes #245.

Good evening! I went through what we already have, and replaced the print statements generated with`fmt` by `log`. However, I noticed two issues, I would like your thoughts on.
* In the _**handlers.go**_ function, replacing the use of `fmt` in this type of statement `c.sendToTg(fmt.Sprintf(partFmt, e.Source.Name))` has not worked with `log` because the later writes to a stream, whereas the statement writes to a char array with fmt. An alternative to `fmt` would be to use `strings.Builder` but I find it more complex than the current statement.

* Adopting levels like _All, Useful_, would require the change of the flag setup from bool to int, and might require passing the debug level argument when calling `go ircClient.StartBot(ircChan, tgClient.SendMessage)` and `go tgClient.StartBot(tgChan, ircClient.SendMessage)`. To my current knowledge of golang, it would mean the use of if-else statements to determine whether to print or not.

### DIFF
--- a/cmd/teleirc.go
+++ b/cmd/teleirc.go
@@ -3,8 +3,9 @@ package main
 
 import (
 	"flag"
-	"fmt"
 	"os"
+	"log"
+	"io/ioutil"
 
 	tgbotapi "github.com/go-telegram-bot-api/telegram-bot-api"
 	"github.com/ritlug/teleirc/internal"
@@ -20,24 +21,30 @@ var (
 	flagPath    = flag.String("conf", ".env", "config file")
 	flagDebug   = flag.Bool("debug", false, "enable debugging")
 	flagVersion = flag.Bool("version", false, "displays current version of TeleIRC")
+
+	logFlags    = log.Ldate | log.Ltime | log.Lmicroseconds | log.Llongfile
+	Info        = log.New(os.Stdout, "INFO: ", logFlags)
+	Warning     = log.New(os.Stdout, "WARNING: ", logFlags)
+	Error       = log.New(os.Stderr, "ERROR: ", logFlags)
+	Ignored     = log.New(ioutil.Discard, "ERROR: ", logFlags)
 )
 
 func main() {
 	flag.Parse()
 
 	if *flagVersion {
-		fmt.Printf("Current TeleIRC version: %s\n", version)
+		Info.Printf("Current TeleIRC version: %s\n", version)
 		return
 	}
 
 	// TODO: Build out debugging capabilities for more verbose output
 	if *flagDebug {
-		fmt.Printf("Debug mode currently set to: %t\n", *flagDebug)
+		Info.Printf("Debug mode currently set to: %t\n", *flagDebug)
 	}
 
 	settings, err := internal.LoadConfig(*flagPath)
 	if err != nil {
-		fmt.Fprintln(os.Stderr, err)
+		Error.Println(err)
 		os.Exit(1)
 	}
 
@@ -52,8 +59,8 @@ func main() {
 
 	select {
 	case ircErr := <-ircChan:
-		fmt.Println(ircErr)
+		Error.Println(ircErr)
 	case tgErr := <-tgChan:
-		fmt.Println(tgErr)
+		Error.Println(tgErr)
 	}
 }
--- a/cmd/teleirc.go
+++ b/cmd/teleirc.go
@@ -24,7 +24,7 @@ var (
 func main() {
 	flag.Parse()
 
-	verbose := internal.Debug {DebugLevel:  *flagDebug	}
+	verbose := internal.Debug { DebugLevel:  *flagDebug }
 
 	if *flagVersion {
 		verbose.PrintVersion("Current TeleIRC version: " + version)
--- a/cmd/teleirc.go
+++ b/cmd/teleirc.go
@@ -27,7 +27,7 @@ func main() {
 	verbose := internal.Debug { *flagDebug	}
 
 	if *flagVersion {
-		verbose.LogInfo("Current TeleIRC version: " + version)
+		verbose.PrintVersion("Current TeleIRC version: " + version)
 		return
 	}
 
--- a/cmd/teleirc.go
+++ b/cmd/teleirc.go
@@ -24,9 +24,7 @@ var (
 
 	logFlags    = log.Ldate | log.Ltime | log.Lmicroseconds | log.Llongfile
 	Info        = log.New(os.Stdout, "INFO: ", logFlags)
-	Warning     = log.New(os.Stdout, "WARNING: ", logFlags)
 	Error       = log.New(os.Stderr, "ERROR: ", logFlags)
-	Ignored     = log.New(ioutil.Discard, "ERROR: ", logFlags)
 )
 
 func main() {
--- a/cmd/teleirc.go
+++ b/cmd/teleirc.go
@@ -24,7 +24,7 @@ var (
 func main() {
 	flag.Parse()
 
-	verbose := internal.Debug { *flagDebug	}
+	verbose := internal.Debug {DebugLevel:  *flagDebug	}
 
 	if *flagVersion {
 		verbose.PrintVersion("Current TeleIRC version: " + version)
--- a/cmd/teleirc.go
+++ b/cmd/teleirc.go
@@ -4,8 +4,6 @@ package main
 import (
 	"flag"
 	"os"
-	"log"
-	"io/ioutil"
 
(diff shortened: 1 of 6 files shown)


Tjzabel:
@10eMyrT in regards to your first comment, the `fmt` here is formatting the string to send over to Telegram. This should not be touched by any logging.

If possible, I would like us to be able to instantiate a `log` object only once or twice, and not in every file the logging is needed. Commonly, this would mean adding a `log` object within our `Client` structs. However, this is a good first start. 

I will add my definitive review once I can get around to testing how this logging works :+1:

internal/handlers/irc/irc.go
```diff
@@ -17,11 +18,20 @@ type Client struct {
 	sendToTg func(string)
 }
 
+/*
+Creating variables for logging
+*/
+var (
+	logFlags    = log.Ldate | log.Ltime | log.Lmicroseconds | log.Llongfile
+	Info        = log.New(os.Stdout, "INFO: ", logFlags)
+	Error       = log.New(os.Stderr, "ERROR: ", logFlags)
```
Zedjones: At the moment, this causes the code not to compile. You need to import `os` as well to use it's namespace. This also applies to telegram.go.

internal/handlers/irc/irc.go
```diff
@@ -17,11 +18,20 @@ type Client struct {
 	sendToTg func(string)
 }
 
+/*
+Creating variables for logging
+*/
+var (
+	logFlags    = log.Ldate | log.Ltime | log.Lmicroseconds | log.Llongfile
```
Zedjones: Could you please switch instances of `Llongfile` to `Lshortfile`? I think having the fully qualified name is most unnecessary here, since most deployments will use something like systemd or containers and the full path wouldn't be very helpful in these situatoins.

internal/handlers/irc/irc.go
```diff
@@ -1,7 +1,8 @@
 package irc
 
 import (
-	"fmt"
+	"log"
+	"io/ioutil"
```
Zedjones: Unused import here, as well as the other files that import this.

Zedjones:
As Tim was saying, it would be a good idea to only have two logger objects (one for error and one for info) and store them in the respective client structs. Or, even better, have a logger struct that contains methods for "logInfo" and "logError" which handles calling the correct logger's `Println` method. 

Also, just running a `go build cmd/teleirc.go` is a good way to make sure that there are no compiler errors/warnings before committing in the future (since we currently don't have any CI set up for Go).

internal/handlers/irc/irc.go
Zedjones: Oh, also, since Go has syntactic capitalization, these variables are now both exported from this module, which is not desired. Please change these to `info` and `error`, respectively.

internal/handlers/irc/irc.go
10eMyrT: Thanks for the comments. I have considered Tim's and your feedback to reorganise. I should have something for today's meeting.

10eMyrT:
I considered the feedback and modified the code. Note that in its current format, the print statements are called in `debug.go`. Therefore, I removed the `Lshortfile` logFlag because it would always print `debug.go`. I also guess knowing the file is not important for INFO. It might be for WARNING and ERROR. There is the [runtime package](https://golang.org/pkg/runtime/) which allows to get the caller function, the filePath would be a long string starting from `/home/......`.

internal/debug.go
```diff
@@ -0,0 +1,42 @@
...
+)
+
+var (
+	// logFlags    = log.Ldate | log.Ltime | log.Lshortfile
+	logFlags    = log.Ldate | log.Ltime
+	info = log.New(os.Stdout, "INFO: ", logFlags)
+	errorLog    = log.New(os.Stderr, "ERROR: ", logFlags)
+	warning     = log.New(os.Stderr, "WARNING: ", logFlags)
+)
+
```
Zedjones: Could you add an interface that defines these functions, something like:
```go
type DebugLogger interface {
	LogInfo(string)
	LogError(error)
	LogWarning(string)
	PrintVersion(string)
}
```
This will help a lot with testing later on.

internal/handlers/irc/irc.go
```diff
@@ -14,14 +13,15 @@ and the IRCSettings that were passed into NewClient
 type Client struct {
 	*girc.Client
 	Settings internal.IRCSettings
+	verbose internal.Debug
```
Zedjones: Once you define that interface, please change structs to contain an `internal.DebugLogger` (or whatever you call it) instead of `internal.Debug`

internal/handlers/irc/irc.go
```diff
@@ -14,14 +13,15 @@ and the IRCSettings that were passed into NewClient
...
 	sendToTg func(string)
 }
 
 /*
 NewClient returns a new IRCClient based on the provided settings
 */
-func NewClient(settings internal.IRCSettings) Client {
-	fmt.Println("Creating new IRC bot client...")
+func NewClient(settings internal.IRCSettings, debug internal.Debug) Client {
```
Zedjones: You'll also need to change the `NewClient` function signatures to accept the interface instead

jwflory:
@Tjzabel @Zedjones Do you think this PR is ready to go?

Zedjones:
> @Tjzabel @Zedjones Do you think this PR is ready to go?

I've just got one last change to request. and it's a linting error. Assuming this has been rebased since the other PR got landed, it should be almost good to go.

cmd/teleirc.go
```diff
@@ -18,42 +17,43 @@ const (
...
 	flagPath    = flag.String("conf", ".env", "config file")
-	flagDebug   = flag.Bool("debug", false, "enable debugging")
+	flagDebug   = flag.Bool("debug", false, "disable debugging")
 	flagVersion = flag.Bool("version", false, "displays current version of TeleIRC")
 )
 
 func main() {
 	flag.Parse()
 
+	verbose := internal.Debug { *flagDebug	}
```
Zedjones: Could you change this to reference the field name? Not a huge deal, but it's a linting error for me.
```go
verbose := internal.Debug{DebugLevel: *flagDebug}
```

10eMyrT: Sure. I already did it.

jwflory:
Awesome. Everything looks good to me. Going to merge this in! :ocean: Thanks for working on this one @10eMyrT. :tada:

Tjzabel:
I just saw these updates. I added my own review of this PR accidentally into #264

PR #262, opened 2020-03-12
ci: Use golangci-lint for Go CI tests (closes #256)

### LINKED ISSUE #256: Add golint to Travis CI tests
_Discussed in [2020-03-07 meeting](https://bluejeans.com/s/jQvmg/)._

---

# Summary

Add [golint](https://github.com/golang/lint) into our Travis CI jobs to require conformance to Go code style on new PRs

# Background

Go code needs to be formatted in a consistent way across our code base.

Golang upstream provides a standard tool to lint Go code, [golint](https://github.com/golang/lint). Our CI should run `golint` on our code to ensure it meets the spec.

I didn't look much into other linting tools, but to keep it simple, it seems like using the official upstream linter is a good place to start.

# Details

Need to make sure `golint` is available in the build `$PATH`. Then, it should be run against the Go code. If it passes the lint check, then the test passes.

# Outcome

When we make new PRs for v2.x.x, Go code is always linted to a common standard

jwflory:
This commit adds the golangci-lint tool to our Go Travis builds. It took
a while to figure out, but this works and successfully lints our code.
It provides some helpful errors when things are not working as they
should be.

At time of making a pull request, the tests are indeed failing. However,
this is helpful feedback for us to iterate on in new PRs. :smile: I
propose merging this through even though the tests are failing so we can
begin using this in all future changes.

Closes #256.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,11 +3,20 @@ os: linux
 dist: bionic
 jobs:
   include:
-    - language: minimal
-      # Set a basic TravisCI file as a placeholder until
-      # we start creating unit tests
-      test: |
-        echo "Hello World"
+    - language: go
+      go: 1.12.x
+      before_install: "curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin $GOLANGCI_LINT_VERSION"
+      script: "golangci-lint run"
+
+    - language: go
+      go: 1.13.x
+      before_install: "curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin $GOLANGCI_LINT_VERSION"
+      script: "golangci-lint run"
+
+    - language: go
+      go: master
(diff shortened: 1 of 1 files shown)


Tjzabel:
@jwflory I really don't like piping into shells and doing all that jazz. Is that really needed?

Would it be simpler to run `go lint`, `go vet` etc within the travis Go environment?

jwflory:

---

We discussed this a bit in the meeting. The official golang/lint tool is not recommended for CI pipelines and even upstream claims that its use is limited:

> The suggestions made by golint are exactly that: suggestions. Golint is not perfect, and has both false positives and false negatives. Do not treat its output as a gold standard. We will not be adding pragmas or other knobs to suppress specific warnings, so do not expect or require code to be completely "lint-free". In short, this tool is not, and will never be, trustworthy enough for its suggestions to be enforced automatically, for example as part of a build process.

That said, the golintci-lint tool is more specifically designed for this use case in mind. It is an unusual and non-standard installation method, but it is supposedly in the name of speed and consistency. I was also +1 to golintci-lint after reading the "Trusted by" part of their README. Even Google uses it:

https://github.com/golangci/golangci-lint#trusted-by

Tjzabel:
@jwflory since this PR is a "failed CI test," can you open an issue to fix the tests?

jwflory:
Rebased this on latest `go-port`. Thanks for the review @Tjzabel. I'll open a new issue to track the failing tests. :+1: 

Merging! :clapper:

PR #263, opened 2020-03-12
ci: Cache Python packages for faster builds

jwflory:
While working on the CI improvements for Go in #262, I discovered a way
to improve the overall speed and performance of the Python job that
builds and tests our documentation. By piping the requirements to `pip`,
we can leverage the caching in Travis CI so it doesn't have to install
the docs dependencies every time a new build is made.

Not sure exactly how fast it will speed things up, but thought it was
helpful to do anyways since it wasn't that hard.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,6 @@
+---
+os: linux
+dist: bionic
 jobs:
   include:
     - language: minimal
@@ -8,8 +11,9 @@ jobs:
 
     - language: python
       python: 3.7
+      cache: pip
       before_install: cd docs/
-      install: pipenv install --dev
+      install: "pip install $(pipenv lock --requirements)"
       script: sh build_docs.sh
 
 notifications:


Tjzabel:
@jwflory my only concern here is that caching can cause all sorts of problems, and we've run into some previously within TeleIRC. Are you able to test this and see how much faster the process speeds up?

jwflory:

---

@Tjzabel We have run into caching issues before with Node, but I am confident in these changes. I've implemented this after seeing examples in other projects and I have added it to other repos already.

Since there aren't many Python packages for the project (three to be exact), it might be an advantage of 15-30 seconds faster. Basically the `pip install` step is skipped unless there are changes to the `Pipfile.lock` file. But since the Python dependencies rarely change, I think the build speed advantage is a useful gain.

jwflory:
Thanks @Tjzabel! In it goes. :clapper:

Issue #248, opened 2020-02-26
Create Document Handler


Tjzabel:
# Summary

TeleIRC v2.0 does not currently support sending documents from Telegram => IRC. 

# Background

**Is your feature request related to a problem? Please describe**:

Documents are not currently picked up by TeleIRC, and IRC users are not notified when a Telegram user has uploaded a document.

**Describe the solution you'd like**:

A document handler needs to be written in order to handle document uploads.

**Describe alternatives you've considered**:

No alternatives.

# Details

This issue is in part related to #159 
The Telegram [document](https://core.telegram.org/bots/api#document) object type will be used.
For completion of this _specific_ issue, the document handler needs to:
1. Document updates should get sent to this handler
2. handler should describe filetype, caption, and user who sent file
3. _**Actual file link should not get sent to IRC (#241)**_

# Outcome

IRC users are able to see someone from Telegram sent a document to the channel.

Tjzabel:
Done via #273

PR #264, opened 2020-03-14
Add Document Handler

10eMyrT:
This is linked to the issue #248 (Creation of the document handler). Currently, it just sends a message notifying that a document is received, but it also contains a variable named Link which will contain a URL, once we decide where to host the documents. The file compiles, but since my environment has not been successfully set, I have not sent a document to test the handler. Please let me know if it does not work in any way.

### DIFF
--- a/cmd/teleirc.go
+++ b/cmd/teleirc.go
@@ -3,7 +3,6 @@ package main
 
 import (
 	"flag"
-	"fmt"
 	"os"
 
 	tgbotapi "github.com/go-telegram-bot-api/telegram-bot-api"
@@ -18,42 +17,43 @@ const (
 
 var (
 	flagPath    = flag.String("conf", ".env", "config file")
-	flagDebug   = flag.Bool("debug", false, "enable debugging")
+	flagDebug   = flag.Bool("debug", false, "disable debugging")
 	flagVersion = flag.Bool("version", false, "displays current version of TeleIRC")
 )
 
 func main() {
 	flag.Parse()
 
+	verbose := internal.Debug { DebugLevel:  *flagDebug }
+
 	if *flagVersion {
-		fmt.Printf("Current TeleIRC version: %s\n", version)
+		verbose.PrintVersion("Current TeleIRC version: " + version)
 		return
 	}
 
 	// TODO: Build out debugging capabilities for more verbose output
-	if *flagDebug {
-		fmt.Printf("Debug mode currently set to: %t\n", *flagDebug)
-	}
+	// Notify that debug is enabled
+	verbose.LogInfo("Debug mode enabled!")
 
 	settings, err := internal.LoadConfig(*flagPath)
 	if err != nil {
-		fmt.Fprintln(os.Stderr, err)
+		verbose.LogError(err)
 		os.Exit(1)
 	}
 
 	var tgapi *tgbotapi.BotAPI
-	tgClient := tg.NewClient(settings.Telegram, tgapi)
+	tgClient := tg.NewClient(settings.Telegram, tgapi, verbose)
 	tgChan := make(chan error)
 
-	ircClient := irc.NewClient(settings.IRC)
+	ircClient := irc.NewClient(settings.IRC, verbose)
 	ircChan := make(chan error)
 	go ircClient.StartBot(ircChan, tgClient.SendMessage)
 	go tgClient.StartBot(tgChan, ircClient.SendMessage)
 
 	select {
 	case ircErr := <-ircChan:
-		fmt.Println(ircErr)
+		verbose.LogError(ircErr)
 	case tgErr := <-tgChan:
-		fmt.Println(tgErr)
+		verbose.LogError(tgErr)
 	}
 }
--- a/internal/config.go
+++ b/internal/config.go
@@ -1,6 +1,7 @@
 package internal
 
 import (
+	"fmt"
 	"os"
 
 	"github.com/caarlos0/env/v6"
@@ -18,7 +19,8 @@ type IRCSettings struct {
 	Port                int      `env:"IRC_PORT" envDefault:"6667" validate:"min=0,max=65535"`
 	TLSAllowSelfSigned  bool     `env:"IRC_CERT_ALLOW_SELFSIGNED" envDefault:"true"`
 	TLSAllowCertExpired bool     `env:"IRC_CERT_ALLOW_EXPIRED" envDefault:"true"`
-	Channel             string   `env:"IRC_CHANNEL,required"`
+	Channel             string   `env:"IRC_CHANNEL,required" validate:"notempty"`
+	ChannelKey          string   `env:"IRC_CHANNEL_KEY" envDefault:""`
 	BotName             string   `env:"IRC_BOT_NAME" envDefault:"teleirc"`
 	SendStickerEmoji    bool     `env:"IRC_SEND_STICKER_EMOJI" envDefault:"true"`
 	SendDocument        bool     `env:"IRC_SEND_DOCUMENT" envDefault:"true"`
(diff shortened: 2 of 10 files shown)


jwflory:

---

I did a walk-through with @10eMyrT on feature branching and `git rebase` after the developer meeting. This PR is blocked, for now, on @Tjzabel's Update handler PR. In the meanwhile, @10eMyrT will work on getting the git history cleaned up on this PR.

10eMyrT:
Apologies on the mess with closing an reopening. I unknowingly closed when I was cleaning the history.

cmd/teleirc.go
```diff
@@ -18,42 +17,43 @@ const (
 
 var (
 	flagPath    = flag.String("conf", ".env", "config file")
-	flagDebug   = flag.Bool("debug", false, "enable debugging")
+	flagDebug   = flag.Bool("debug", false, "disable debugging")
```
Tjzabel: I think keeping `enable debugging` is better, since this lets users know that any changes to this `Bool` adds debugging capabilities.

internal/config.go
```diff
@@ -18,7 +19,8 @@ type IRCSettings struct {
 	Port                int      `env:"IRC_PORT" envDefault:"6667" validate:"min=0,max=65535"`
 	TLSAllowSelfSigned  bool     `env:"IRC_CERT_ALLOW_SELFSIGNED" envDefault:"true"`
 	TLSAllowCertExpired bool     `env:"IRC_CERT_ALLOW_EXPIRED" envDefault:"true"`
-	Channel             string   `env:"IRC_CHANNEL,required"`
+	Channel             string   `env:"IRC_CHANNEL,required" validate:"notempty"`
+	ChannelKey          string   `env:"IRC_CHANNEL_KEY" envDefault:""`
```
Tjzabel: This seems to be a part of another PR?

internal/debug.go
```diff
@@ -0,0 +1,50 @@
...
+*/
+package internal
+
+import (
+	"log"
+	"os"
+)
+
+var (
+	// logFlags    = log.Ldate | log.Ltime | log.Lshortfile
```
Tjzabel: This comment seems like it can be removed

internal/debug.go
```diff
@@ -0,0 +1,50 @@
...
+}
+
+func (d Debug) LogInfo(message string) {
+	if d.DebugLevel {
+		info.Println(message)
+	}
+}
+
+func (d Debug) LogError(message error) {
+	errorLog.Println(message)
```
Tjzabel: Does `Println` take into account the type of output stream to use?

internal/handlers/telegram/documentHandler.go
```diff
@@ -0,0 +1,39 @@
+package telegram
+
+import (
+	tgbotapi "github.com/go-telegram-bot-api/telegram-bot-api"
+)
+
+
+type DocumentObject struct {
+	EnableLink bool
```
Tjzabel: WIth the v2.0 iteration of TeleIRC, I would like to move away from enabling any sort of boolean for sending documents.

internal/handlers/telegram/documentHandler.go
```diff
@@ -0,0 +1,39 @@
...
+	Link string
+	Object *tgbotapi.Document
```
Tjzabel: Instead of calling this Object, we can call this something a little nicer like `Document`.

internal/handlers/telegram/documentHandler.go
```diff
@@ -0,0 +1,39 @@
...
+}
+
+/*
+Constructs a new document and is used to decide under which format to formard
```
Tjzabel: This relates to the issue we have up on how to do proper development documentation. GoDoc uses comments to generate dev docs. Start function comments with "functionName does...."

In this case, we can start off with "newDocument constructs a new document and is..."

internal/handlers/telegram/documentHandler.go
```diff
@@ -0,0 +1,39 @@
...
+	return DocumentObject{enabled, link, doc}
+}
+
+/*
+documentHandler once a document received, uses information about the document
+to format a message to return to the client.
+*/
+func (doc DocumentObject) documentHandler()  string {
+
+	message := " shared a file (" +
```
Tjzabel: We need to display the Telegram username as well. This doesn't tell us _who_ shared the file.

internal/handlers/telegram/documentHandler.go
```diff
@@ -0,0 +1,39 @@
...
+		doc.Object.MimeType + ") on Telegram with title" +
+		"'" + doc.Object.FileName + "'."
+
+	if doc.EnableLink {
```
Tjzabel: As I mentioned earlier, I think we should remove the boolean from this.

internal/handlers/telegram/documentHandler.go
```diff
@@ -0,0 +1,39 @@
...
+		message += " View document at " + doc.Link
```
Tjzabel: Since we're only sending over that someone sent a document in Telegram, we don't need to include the link here.

cmd/teleirc.go
```diff
@@ -18,42 +17,43 @@ const (
...
 
+	verbose := internal.Debug { DebugLevel:  *flagDebug }
+
 	if *flagVersion {
-		fmt.Printf("Current TeleIRC version: %s\n", version)
+		verbose.PrintVersion("Current TeleIRC version: " + version)
 		return
 	}
 
 	// TODO: Build out debugging capabilities for more verbose output
```
Tjzabel: We can remove this comment.

cmd/teleirc.go
```diff
@@ -18,42 +17,43 @@ const (
...
 	flagPath    = flag.String("conf", ".env", "config file")
-	flagDebug   = flag.Bool("debug", false, "enable debugging")
+	flagDebug   = flag.Bool("debug", false, "disable debugging")
 	flagVersion = flag.Bool("version", false, "displays current version of TeleIRC")
 )
 
 func main() {
 	flag.Parse()
 
+	verbose := internal.Debug { DebugLevel:  *flagDebug }
```
Tjzabel: I wonder if it's better to just call this `log` or `debug` instead of verbose, so it's explicitly clear this is part of the logging.
I can go either way with this, just a thought.

internal/debug.go
```diff
@@ -0,0 +1,50 @@
+/*
+	Debug contains information about debug levels and define
```
Tjzabel: Not sure why the comments here are indented

internal/debug.go
```diff
@@ -0,0 +1,50 @@
...
+
+import (
+	"log"
+	"os"
+)
+
+var (
+	// logFlags    = log.Ldate | log.Ltime | log.Lshortfile
+	logFlags    = log.Ldate | log.Ltime
+	info = log.New(os.Stdout, "INFO: ", logFlags)
```
Tjzabel: This should be indented properly with the rest

internal/handlers/irc/irc_test.go
```diff
@@ -36,7 +39,10 @@ func TestNewClientFull(t *testing.T) {
 		BotName:          "test_name",
 		NickServPassword: "test_pass",
 	}
-	client := NewClient(ircSettings)
+	logger := internal.Debug{
+		DebugLevel:  false,
+	}
+	client := NewClient(ircSettings, logger)
```
Tjzabel: Do these tests pass?

Tjzabel:
_**EDIT**_:  It looks like this PR encompasses 2-3 others. I've added feedback for both the logging PR and this document PR into this one. It seems I was not able to review the logging PR before it was merged in.

internal/handlers/telegram/documentHandler.go
10eMyrT: Sure, I like the idea.

internal/handlers/irc/irc_test.go
10eMyrT: Yes, all the tests pass.

internal/config.go
10eMyrT: Yes, it is. In this PR, I have not changed anything in this file.

internal/debug.go
10eMyrT: `Println` simply prints the error message to stdout, as it does with `fmt`. Is there an additional need?

Tjzabel: @10eMyrT we may want this to print out to standard error, not standard out.

10eMyrT: I see. That's the best way. I will integrate it in the changes before next meeting.

jwflory:
_Discussed in [2020-03-21 meeting](https://bluejeans.com/s/3YKnj/)._

---

@10eMyrT is going to split this up into two smaller PRs from separate feature branches: one for the logging changes and one for the document handler.

Issue #266, opened 2020-03-15
Fix unused vars and unchecked returns


jwflory:
# Summary

Make CI tests pass again by fixing unused variables and unchecked returns

# Standard debugging steps

## How to reproduce?

Reproduced in our CI pipeline. See below:

https://travis-ci.org/github/RITlug/teleirc/jobs/662793390#L271

## Expected behavior

CI tests should pass.

## Stacktrace

```go
internal/handlers/telegram/handler.go:8:2: `joinFmt` is unused (varcheck)
	joinFmt = "%s (@%s) has joined the Telegram Group!"
	^
internal/handlers/telegram/handler.go:9:2: `partFmt` is unused (varcheck)
	partFmt = "%s (@%s) has left the Telegram Group."
	^
internal/config.go:99:29: Error return value of `validate.RegisterValidation` is not checked (errcheck)
	validate.RegisterValidation("notempty", validateEmptyString)
	                           ^
internal/handlers/telegram/telegram.go:35:13: Error return value of `tg.api.Send` is not checked (errcheck)
	tg.api.Send(newMsg)
	           ^
```

Zedjones:
In regards to this error:
```go
internal/handlers/telegram/telegram.go:35:13: Error return value of `tg.api.Send` is not checked (errcheck)
	tg.api.Send(newMsg)
	           ^
``` 
It's a bit more complicated than might be initially obvious. During the call on 3/21, we agree that it would be best to check `tg.api.Send` for the error, log it, and retry sending a few times before giving up.

PR #268, opened 2020-03-21
Add timeout to girc connect call

### LINKED ISSUE #253: Add timeout to girc connect call
# Summary

Currently, if `girc.Connect` is called and the server is reached but never responds, then the goroutine will just hang forever. Using `girc.DialerConnect` instead, we can add a custom dialer which includes a timeout when connecting to the server.

# Background

`girc.Connect` can currently hang forever

We should add a timeout with a custom dialer.

# Details

# Outcome

Zedjones:
Before this gets merged, I want to get everybody's feedback on whether or not the timeout should be configurable. I set it to 5 seconds for now, but there's no particularly reason, I just chose that time. It can be adjusted if we decide to keep it hard-coded.

Closes #253

### DIFF
--- a/internal/handlers/irc/irc.go
+++ b/internal/handlers/irc/irc.go
@@ -47,9 +47,8 @@ func (c Client) StartBot(errChan chan<- error, sendMessage func(string)) {
 	c.logger.LogInfo("Starting up IRC bot...")
 	c.sendToTg = sendMessage
 	c.addHandlers()
-	// TODO: Currently just set to 5 seconds,
-	// if we want this to be configurable we can add a config option
-	if err := c.DialerConnect(&net.Dialer{Timeout: 5 * time.Second}); err != nil {
+	// 10 second timeout for connection
+	if err := c.DialerConnect(&net.Dialer{Timeout: 10 * time.Second}); err != nil {
 		errChan <- err
 		c.logger.LogError(err)
 	} else {
--- a/internal/handlers/irc/irc.go
+++ b/internal/handlers/irc/irc.go
@@ -15,7 +15,7 @@ and the IRCSettings that were passed into NewClient
 type Client struct {
 	*girc.Client
 	Settings internal.IRCSettings
-	verbose  internal.DebugLogger
+	logger   internal.DebugLogger
 	sendToTg func(string)
 }
 
@@ -44,14 +44,14 @@ StartBot adds necessary handlers to the client and then connects,
 returns any errors that occur
 */
 func (c Client) StartBot(errChan chan<- error, sendMessage func(string)) {
-	c.verbose.LogInfo("Starting up IRC bot...")
+	c.logger.LogInfo("Starting up IRC bot...")
 	c.sendToTg = sendMessage
 	c.addHandlers()
 	// TODO: Currently just set to 5 seconds,
 	// if we want this to be configurable we can add a config option
 	if err := c.DialerConnect(&net.Dialer{Timeout: 5 * time.Second}); err != nil {
 		errChan <- err
-		c.verbose.LogError(err)
+		c.logger.LogError(err)
 	} else {
 		errChan <- nil
 	}
@@ -70,7 +70,7 @@ addHandlers adds handlers for the client struct based on the settings
 that were passed in to NewClient
 */
 func (c Client) addHandlers() {
-	c.verbose.LogInfo("Adding IRC event handlers...")
+	c.logger.LogInfo("Adding IRC event handlers...")
 	for eventType, handler := range getHandlerMapping() {
 		c.Handlers.Add(eventType, handler(c))
 	}
--- a/internal/handlers/irc/irc.go
+++ b/internal/handlers/irc/irc.go
@@ -1,6 +1,8 @@
 package irc
 
 import (
+	"net"
+	"time"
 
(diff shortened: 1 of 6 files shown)


jwflory:
@Zedjones In favor of end-user simplicity, I think we can use a sane hard-coded default. If we get feedback from users about timeout issues, we can either extend it or make it configurable in the future.

For now, I prefer to ship a hard-coded default to keep our config simple.

internal/handlers/irc/irc.go
```diff
@@ -13,7 +15,7 @@ and the IRCSettings that were passed into NewClient
 type Client struct {
 	*girc.Client
 	Settings internal.IRCSettings
-	verbose internal.DebugLogger
+	verbose  internal.DebugLogger
```
jwflory: Not sure if you meant to add extra whitespace here :joy:

10eMyrT: That was on my end with the logging.

internal/handlers/telegram/handler.go
```diff
@@ -4,10 +4,13 @@ import (
 	tgbotapi "github.com/go-telegram-bot-api/telegram-bot-api"
 )
 
+// TODO: These format strings are currently unused, so commenting out for now
```
Tjzabel: Thanks for clarifying these aren't yet used :+1:

internal/handlers/irc/irc.go
```diff
@@ -45,7 +47,9 @@ func (c Client) StartBot(errChan chan<- error, sendMessage func(string)) {
 	c.verbose.LogInfo("Starting up IRC bot...")
 	c.sendToTg = sendMessage
 	c.addHandlers()
-	if err := c.Connect(); err != nil {
+	// TODO: Currently just set to 5 seconds,
+	// if we want this to be configurable we can add a config option
+	if err := c.DialerConnect(&net.Dialer{Timeout: 5 * time.Second}); err != nil {
```
Tjzabel: I'm struggling to think of whether or not to add this to our config. Personally, I think this should be set to something like 10 seconds to account for any delays IRC side.

internal/handlers/irc/irc.go
```diff
@@ -45,7 +47,9 @@ func (c Client) StartBot(errChan chan<- error, sendMessage func(string)) {
 	c.verbose.LogInfo("Starting up IRC bot...")
 	c.sendToTg = sendMessage
 	c.addHandlers()
-	if err := c.Connect(); err != nil {
+	// TODO: Currently just set to 5 seconds,
```
Tjzabel: I would like these comments to be removed before this PR merges, once we settle on a time.

internal/handlers/irc/irc.go
Zedjones: Yeah, gofmt did that to make it so that the definitions line up haha. Normally this stuff sorta vanishes in PRs, but I don't think Kevin's setup is doing formatting so that's why it didn't show up until this PR.

internal/handlers/irc/irc.go
jwflory: @Tjzabel Based on our design goals for v2.0.0, I'd prefer to err on the side of simplicity and leave it out, since I think most cases won't need this to be a custom value. If we are wrong and it turns out it needs to be configurable, it shouldn't be that hard to convert later.

internal/handlers/irc/irc.go
Zedjones: Yeah, that was the plan 👍

jwflory: I'd like to keep the comment that states it as a 5 second timeout but we can trim the second line

internal/handlers/irc/irc.go
Zedjones: Do we want to move it to 10 seconds, then?

internal/handlers/irc/irc.go
jwflory: Oh okay! No problem, let's keep it.

internal/handlers/irc/irc.go
Tjzabel: @jwflory since `time` libraries are in every single language, I'm not sure we even need the 5 seconds clarification comment

jwflory: I am +0 on this one, not going to block on it :slightly_smiling_face:

internal/handlers/irc/irc.go
jwflory: Sure, 10 seconds works for me.

jwflory:
This PR is ready for a rebase now that #274 is merged.

10eMyrT:
Conflict solved. It is ready for merging.

jwflory:
Thanks @10eMyrT, you are becoming a pro at git rebasing and merge conflicts! :smile: 

I think we are still waiting on a bump from 5 seconds timeout to 10 seconds. Once @Zedjones has a chance to change that, we can merge this in. I'm getting ready for the Mega Merge up next! :grin:

10eMyrT:
Right, we still have decide on that. I have finished with #266, just waiting for this PR to avoid conflicts with the one I have.

jwflory:
Awesome. Thanks everyone! Going to merge this in. :soccer: :arrow_right: :goal_net:

PR #273, opened 2020-03-21
Create document handler

### LINKED ISSUE #248: Create Document Handler

10eMyrT:
In this PR, the document handler function is added to `handler.go` in order to maintain all handlers together in the same file.

Closes #248.

### DIFF
--- a/internal/handlers/telegram/handler.go
+++ b/internal/handlers/telegram/handler.go
@@ -26,3 +26,20 @@ func messageHandler(tg *Client, u tgbotapi.Update) {
 
 	tg.sendToIrc(formatted)
 }
+
+/*
+documentHandler receives a document object from Telegram, and sends
+a notification to IRC.
+*/
+func documentHandler(tg *Client, u *tgbotapi.Message) {
+
+	formatted := u.From.UserName + " shared a file (" + u.Document.MimeType + ")"
+
+	if u.Caption != "" {
+		formatted += " on Telegram with caption: " + "'" + u.Caption + "'."
+	} else {
+		formatted += " on Telegram with title: " + "'" + u.Document.FileName + "'."
+	}
+
+	tg.sendToIrc(formatted)
+}
--- a/internal/handlers/telegram/telegram.go
+++ b/internal/handlers/telegram/telegram.go
@@ -65,7 +65,12 @@ func (tg *Client) StartBot(errChan chan<- error, sendMessage func(string)) {
 			continue
 		}
 
-		messageHandler(tg, update)
+		if update.Message.Document != nil {
+			documentHandler(tg, update.Message)
+		} else {
+			messageHandler(tg, update)
+		}
+
 	}
 	errChan <- nil
 }


internal/handlers/telegram/handler.go
```diff
@@ -26,3 +26,17 @@ func messageHandler(tg *Client, u tgbotapi.Update) {
 
 	tg.sendToIrc(formatted)
 }
+
+/*
+documentHandler once a document received, uses information about the document
+to format a message to return to the client.
+*/
+func documentHandler(tg *Client, u tgbotapi.Update) {
+	header := tg.Settings.Prefix + u.Message.From.UserName +
```
Tjzabel: When sending over things like documents and photos, a user isn't strictly saying the message, so we don't need the message prefix and suffix. Here's what the message format looks like on NodeJS:
```
T​jzabel shared a file (image/jpeg) on Telegram with caption: 'my file'
```

internal/handlers/telegram/handler.go
```diff
@@ -26,3 +26,17 @@ func messageHandler(tg *Client, u tgbotapi.Update) {
...
+		tg.Settings.Suffix
+
+	message := " shared a file (" + u.Message.Document.MimeType +
+		") on Telegram with title" + "'" + u.Message.Document.FileName + "'."
```
Tjzabel: I would generally want the caption a user adds to their document, as this would give IRC users more background information as to what the file is. It may be worthwhile sending over the filename instead of a caption if the user does not supply one.

Tjzabel: Additionally, line 39 does not have a space after "with title...", and so the filename is sent over without any spaces.

internal/handlers/telegram/handler.go
```diff
@@ -26,3 +26,17 @@ func messageHandler(tg *Client, u tgbotapi.Update) {
 
 	tg.sendToIrc(formatted)
 }
+
+/*
+documentHandler once a document received, uses information about the document
```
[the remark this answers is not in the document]
10eMyrT: The comment you proposed looks better to me. I used yours instead of coming up with a new one.

internal/handlers/telegram/handler.go
10eMyrT: Good eye for the space. I solved it. As of the caption, I used an if-else statement to check if there is a caption. I use the caption if it is not an empty string, and the filename if the caption is empty.

internal/handlers/telegram/handler.go
10eMyrT: Prefix and suffix removed.

PR #274, opened 2020-03-21
Small changes to the debug file and the logging variables

10eMyrT:
This PR considers the comments made by @Tjzabel and made changes to the `debug.go` file as well as the names of the debug variables in other functions.

### DIFF
--- a/cmd/teleirc.go
+++ b/cmd/teleirc.go
@@ -24,36 +24,35 @@ var (
 func main() {
 	flag.Parse()
 
-	verbose := internal.Debug { DebugLevel:  *flagDebug }
+	debug := internal.Debug { DebugLevel:  *flagDebug }
 
 	if *flagVersion {
-		verbose.PrintVersion("Current TeleIRC version: " + version)
+		debug.PrintVersion("Current TeleIRC version: " + version)
 		return
 	}
 
-	// TODO: Build out debugging capabilities for more verbose output
 	// Notify that debug is enabled
-	verbose.LogInfo("Debug mode enabled!")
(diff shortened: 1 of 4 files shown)


cmd/teleirc.go
```diff
@@ -24,36 +24,35 @@ var (
...
-		verbose.LogError(err)
+		debug.LogError(err)
 		os.Exit(1)
 	}
 
 	var tgapi *tgbotapi.BotAPI
-	tgClient := tg.NewClient(settings.Telegram, tgapi, verbose)
-	tgChan := make(chan error)
+	tgClient := tg.NewClient(settings.Telegram, tgapi, debug)
+	tgChan   := make(chan error)
```
Tjzabel: The `:=` on line 45 doesn't need to be in line with line 44. Padding is typically stylistically done when there are 3-4+ variables all being instantiated at the same time.

cmd/teleirc.go
```diff
@@ -24,36 +24,35 @@ var (
...
 
-	ircClient := irc.NewClient(settings.IRC, verbose)
-	ircChan := make(chan error)
+	ircClient := irc.NewClient(settings.IRC, debug)
+	ircChan   := make(chan error)
```
Tjzabel: Same here. Doesn't need to be in line.

10eMyrT:
The feedback from @Tjzabel has been addressed. Thank you @Tjzabel and @jwflory for the review. We can merge then.

jwflory:
@10eMyrT Super, thanks for updating. In it goes! :soccer: :arrow_right: :goal_net:

PR #275, opened 2020-03-27
Fixing unused vars and unchecked errors

10eMyrT:
This PR comes to fix the Travis CI issue. Two files (`handler.go` and `config.go`) have already been addressed by @Zedjones. The main focus is with `tg.api.Send(newMsg)` in `telegram.go`. I will solve the merge conflicts before the merge.

### DIFF
--- a/internal/handlers/telegram/telegram.go
+++ b/internal/handlers/telegram/telegram.go
@@ -32,7 +32,16 @@ SendMessage sends a message to the Telegram channel specified in the settings
 func (tg *Client) SendMessage(msg string) {
 	newMsg := tgbotapi.NewMessage(tg.Settings.ChatID, "")
 	newMsg.Text = msg
-	tg.api.Send(newMsg)
+
+	if _, err := tg.api.Send(newMsg); err != nil {
+		var attempts int = 0
+		// Try resending 3 times if the message is successfully sent
+		for err != nil && attempts < 3 {
+			tg.logger.LogError(err)
+			attempts++
+			_, err = tg.api.Send(newMsg)
+		}
+	}
 }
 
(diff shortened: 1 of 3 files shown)


internal/handlers/telegram/telegram.go
```diff
@@ -32,7 +32,14 @@ SendMessage sends a message to the Telegram channel specified in the settings
...
 	newMsg := tgbotapi.NewMessage(tg.Settings.ChatID, "")
 	newMsg.Text = msg
-	tg.api.Send(newMsg)
+
+	if _, err := tg.api.Send(newMsg); err != nil {
+		// Keep trying until the message is successfully sent
+		for err != nil {
+			tg.logger.LogError(err)
+			_, err = tg.api.Send(newMsg)
+		}
```
10eMyrT: This loop forces to keep sending as long as the message did not go through. I thought about setting a number of trials before moving on, and would suggest ***resending 3 times and move on if it still does not go through.***

10eMyrT:
Conflicts resolved, and Travis checks are successful. We just have to agree on the number of times to try resending.

internal/handlers/telegram/telegram.go
jwflory: @10eMyrT I think three attempts is fair. As long as each failed attempt is logged, which it looks like it is to me.

10eMyrT: I do too. I have changed the code to consider resending 3 times. Let's wait for other inputs before we merge.

10eMyrT:
Thanks @jwflory and @Tjzabel for the reviews. I will merge now.

PR #280, opened 2020-03-28
fmt/log.Println like logging, Adding more verbosity 

kennedy:
1. Increased verbosity:
    - Whenever an irc handler triggers
    - Whenever an irc event handler has been added. Provides more details into which handlers were added.
1. Added log notification when telegram bot has logged in.

### DIFF
--- a/internal/debug.go
+++ b/internal/debug.go
@@ -14,10 +14,10 @@ var (
 
 // DebugLogger provides an interface to call the logging functions
 type DebugLogger interface {
-	LogInfo(string)
-	LogError(error)
-	LogWarning(string)
-	PrintVersion(string)
+	LogInfo(v ...interface{})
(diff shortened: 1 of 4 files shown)


internal/debug.go
```diff
@@ -14,33 +14,33 @@ var (
 
 // DebugLogger provides an interface to call the logging functions
 type DebugLogger interface {
```
kennedy: Behaves just like `fmt.Println()` or `log.Println()`
for example: `logger.LogInfo("random pre-fix string", 10) # random pre-fix string 10`

kennedy:
> 
> 
> would you add a space before concatenating string on these lines?

Fortunately a space is automatically joined between the interfaces

kennedy:
mergeeeed

PR #285, opened 2020-03-29
Create Telegram update handler

### LINKED ISSUE #238: Create Update Handler
## Summary

In order to correctly assign Telegram Updates to their corresponding handlers, we need a top-level `updateHandler` to figure out what type of TG update is coming through, and properly hand off the update to the correct handler.

## Telegram Update Handler Overview

1. Receive an [Update](https://github.com/go-telegram-bot-api/telegram-bot-api/blob/master/types.go#L29)
2. Parse the [Message](https://github.com/go-telegram-bot-api/telegram-bot-api/blob/master/types.go#L133) type
    * If not a `Message` object, check for other `Update` types
3. Hand off action to the correct handler

Tjzabel:
This PR piggy-backs off @kennedy's work and moves the logic for determining which handler to use for a given Telegram update into its own function in the `handler.go` file.

Any new handlers need to have a switch statement added into the `updateHandler` in order to properly route Telegram updates.

Closes #238

### DIFF
--- a/internal/handlers/telegram/handler.go
+++ b/internal/handlers/telegram/handler.go
@@ -4,20 +4,33 @@ import (
 	tgbotapi "github.com/go-telegram-bot-api/telegram-bot-api"
 )
 
-// TODO: These format strings are currently unused, so commenting out for now
-/*
-const (
-	joinFmt = "%s (@%s) has joined the Telegram Group!"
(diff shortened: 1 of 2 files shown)


jwflory:
Hooray! Happy to see this make it in. :smile:

PR #290, opened 2020-03-29
Adding basic LogDebug into the updateHandler

kennedy:
If `--debug=true` flag is provided in startup, telegram update notifications will show as debug logs.

### DIFF
--- a/internal/handlers/telegram/handler.go
+++ b/internal/handlers/telegram/handler.go
@@ -19,14 +19,19 @@ func updateHandler(tg *Client, updates tgbotapi.UpdatesChannel) {
 	for u := range updates {
 		switch {
 		case u.Message == nil:
+			tg.logger.LogDebug("Missing message data")
 			continue
 		case u.Message.Text != "":
+			tg.logger.LogDebug("messageHandler triggered")
 			messageHandler(tg, u)
 		case u.Message.Sticker != nil:
+			tg.logger.LogDebug("stickerHandler triggered")
 			stickerHandler(tg, u)
 		case u.Message.Document != nil:
+			tg.logger.LogDebug("documentHandler triggered")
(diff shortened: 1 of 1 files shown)


internal/handlers/telegram/handler.go
```diff
@@ -19,14 +19,19 @@ func updateHandler(tg *Client, updates tgbotapi.UpdatesChannel) {
...
+			tg.logger.LogDebug("messageHandler triggered")
 			messageHandler(tg, u)
 		case u.Message.Sticker != nil:
+			tg.logger.LogDebug("stickerHandler triggered")
 			stickerHandler(tg, u)
 		case u.Message.Document != nil:
+			tg.logger.LogDebug("documentHandler triggered")
 			documentHandler(tg, u.Message)
 		default:
+			tg.logger.LogDebug("triggered, but message type is currently unsupported")
```
kennedy: I think i will turn this to a warning

internal/handlers/telegram/handler.go
```diff
@@ -19,14 +19,19 @@ func updateHandler(tg *Client, updates tgbotapi.UpdatesChannel) {
 	for u := range updates {
 		switch {
 		case u.Message == nil:
+			tg.logger.LogDebug("Missing message data")
```
kennedy: I think i will turn this to an error

jwflory:
> Looks good to me. In higher levels of debug, I would like to see the actual contents of the messages as well, as this may be something end users will want to log.
> 
> @jwflory what are your thoughts on these logging levels?

@Tjzabel I agree the actual logs of the messages themselves will be helpful, but I also agree it makes sense at a higher level than debugging. This change will be useful if messages aren't appearing in IRC or Telegram, for instance.

I'm going to merge this in. :clapper: Thanks @kennedy!

PR #294, opened 2020-04-02
Telegram settings checks in IRC client

pomo-mondreganto:
I noticed that ShowJoinMessage and ShowLeaveMessage are not used anywhere, and such messages are really trashing my chat, so I added basic checks of those variables in irc events handlers. 

I tested it in my environment, but didn't have enough time to write unit tests (if necessary, I could write them later).

Also, worth mentioning: it's not obvious where the ShowKickMessage should be used.

### DIFF
--- a/internal/handlers/irc/irc.go
+++ b/internal/handlers/irc/irc.go
@@ -14,15 +14,16 @@ and the IRCSettings that were passed into NewClient
 */
 type Client struct {
 	*girc.Client
-	Settings internal.IRCSettings
-	logger   internal.DebugLogger
-	sendToTg func(string)
+	Settings         *internal.IRCSettings
+	TelegramSettings *internal.TelegramSettings
+	logger           internal.DebugLogger
+	sendToTg         func(string)
 }
 
 /*
 NewClient returns a new IRCClient based on the provided settings
 */
-func NewClient(settings internal.IRCSettings, logger internal.DebugLogger) Client {
+func NewClient(settings *internal.IRCSettings, telegramSettings *internal.TelegramSettings, logger internal.DebugLogger) Client {
 	logger.LogInfo("Creating new IRC bot client...")
 	client := girc.New(girc.Config{
 		Server: settings.Server,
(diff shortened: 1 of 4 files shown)


jwflory:
Hey @pomo-mondreganto, thanks for opening up a PR! You are our first contributor to the upcoming v2.0.0 from outside the core developer team. :slightly_smiling_face: 

Someone from @RITlug/teleirc-developers-commit-access will follow up here soon to review your changes. As for the two CodeClimate failed checks, that's on the core team to fix so don't worry about that.

internal/handlers/irc/irc.go
```diff
@@ -14,15 +14,16 @@ and the IRCSettings that were passed into NewClient
 */
 type Client struct {
 	*girc.Client
-	Settings internal.IRCSettings
-	logger   internal.DebugLogger
-	sendToTg func(string)
+	Settings         *internal.IRCSettings
```
Tjzabel, line 17: Thanks for changing these to pointers :+1:

Tjzabel:
@jwflory side note, is there something we can do to turn down the CC checks for the duplicate code? The join/part handlers will indeed be almost duplicates, but I don't wish to join these together.

jwflory:
> @jwflory side note, is there something we can do to turn down the CC checks for the duplicate code? The join/part handlers will indeed be almost duplicates, but I don't wish to join these together.

We can mark checks as `wontfix`. I just did that on the two errors from CodeClimate, so the CodeClimate tests are passing now. :+1: Just wanted to get a confirmation before doing that. I also think others can do that too so long as you log in with your GitHub account to CodeClimate.

Tjzabel:
@jwflory cool, thanks :100:

pomo-mondreganto:
@Tjzabel glad I could help!

Tjzabel:
@Zedjones I would like you to still review this PR (since you mostly work on the IRC-side), but this work now directly impacts my work with the join/part handlers on the Telegram side. Going to merge this through so I can get my fixes in :ocean:

PR #296, opened 2020-04-04
gitignore: Cover more tools and editors

jwflory:
This commit greatly expands the coverage of the `.gitignore` file to the
following types of categories:

* TeleIRC-specific files
* Operating Systems (Linux, macOS, Windows)
* Programming Languages (Go, Python)
* Text Editors (Sublime Text, Vim, VSCode)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,20 +1,7 @@
-# ---------------------------------- Docker ------------------------------------
+############################# ----- TELEIRC ----- #############################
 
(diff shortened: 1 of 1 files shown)


jwflory:
Also I dropped the `Dockerfile` ignore because based on today's dev meeting discussion, we will probably ship just one container image, not multiple like we did in v1.x.x.

PR #289, opened 2020-03-29
Testing telegram message and sticker handlers

kennedy:
For each test per handler, I created a client and update object and pass that through. Is this what we are looking for?

resolves #272
resolves #287 

### passing
```
=== RUN   ExampleStickerHandler
--- PASS: ExampleStickerHandler (0.00s)
=== RUN   ExampleMessageHandler
--- PASS: ExampleMessageHandler (0.00s)
PASS

Process finished with exit code 0
```

### failing
```
=== RUN   ExampleMessageHandler
--- FAIL: ExampleMessageHandler (0.00s)
got:
<test>Random Text
want:
<test>Random wrong Text
FAIL

Process finished with exit code 1
```

Issue #241, opened 2020-02-09
[Security] Telegram Bot Hash Leak


pragmatic-strain:
# Summary

Data (bot hash) leak at IRC channel.

Example:
```
<teleirc_irc_bot_nick> 'Untitled Document' uploaded by telegram_user:
<teleirc_irc_bot_nick> https://api.telegram.org/file/bot523281407:AAGgJwVGwhX7XkQ-5SUKGahXe-yg6QfNWQU/animations/file_0.mp4
```
(Note: The above example of a bot hash was replaced and is no longer valid.)

## How to reproduce?

1. set `IRC_SEND_DOCUMENT=true` at `.env`
2. run teleirc
3. send video/mp4 to telegram group
4. See how teleirc leaks the Telegram bot hash to the irc channel

## Expected behavior

Should not leak bot hash under any circumstances. Please fix as you wish.

Tjzabel:
First off, thanks for the issue :+1: 

## Background
This is actually an issue with the [Telegram Bot API](https://core.telegram.org/bots/api) itself. 
We (TeleIRC) disable sending over documents by default. You can see more history [here](https://github.com/RITlug/teleirc/issues/115).

## Potential Options
There are a few options I can think of:
1. Completely disable sending over documents
2. Look for an intermediary document hosting service like we do with Imgur
3. Reach out to Telegram and see if they have interest in changing this functionality
   * This would most likely break most bots out there that handle documents

kennedy:
Not sure if this is helpful, but I found the API doc for uploading and downloading files for telegram: https://core.telegram.org/api/files

jwflory:
Hi @pragmatic-strain, thanks for opening this. We won't address this for the v1.x.x release, but I opened a new issue (#242) for figuring out a better way to handle this in the v2.x.x release. I'm going to close this issue and we can keep up the discussion in #242.

> Not sure if this is helpful, but I found the API doc for uploading and downloading files for telegram: https://core.telegram.org/api/files

Thanks @kennedy, I edited this into the new issue! :+1:

Issue #159, opened 2019-08-12
Consider using an intermediary document hosting service


Tjzabel:
Due to TeleIRC disabling sending over documents by default due to Telegram sending over the API key, it may be worthwhile looking into possibly using an intermediary measure to translate the Telegram link to another hosting source. 

Another *possible* measure would be to ask Telegram to change how bots send documents. This would fix many issues that I'm sure many other Telegram bots suffer from.

jwflory:
Thanks for opening this @Tjzabel.

> Due to TeleIRC disabling sending over documents by default due to Telegram sending over the API key, it may be worthwhile looking into possibly using an intermediary measure to translate the Telegram link to another hosting source.

We need more research on our options. We could do something fancy like integrate with Google Drive, but that type of storage seems more permanent than Telegram messages are meant for. Do you have some time to explore other ephemeral options?

> Another _possible_ measure would be to ask Telegram to change how bots send documents. This would fix many issues that I'm sure many other Telegram bots suffer from.

The upstream fix is the best but also hardest fix. Despite Telegram touting its open source roots, they do not have a transparent or accessible development community or place to share feedback.

jwflory:
Someone popped into our developer chat and suggested Firefox Send as one possible option for this service. If it is possible to register an API key to send files through an API, this would be a nifty way of doing it. For example, maybe we could set files to expire after a week or a month, or offer that as a config option to the user.

Not sure if this is possible, but from a privacy point-of-view, Firefox Send would be a cool tool to integrate with here.

PR #298, opened 2020-04-17
Testing for the document handler

### LINKED ISSUE #286: Write unit tests for the document handler
# Summary

Unit tests do not currently exist for the Telegram document handler.

# Background

After writing this handler, it is unclear if we have handled all edge cases. We should have unit tests to thoroughly test this functionality.

Full set of unit tests to catch as many edge cases as we can.

No alternatives.

# Details

The owner of this issue needs to think about what a user can and cannot do when uploading a file to Telegram. Then, build unit tests around the desired and undesired functionality to make sure error handling exists in the appropriate locations.

# Outcome

Unit testing for the TG document handler. Allows us to feel safe knowing this handler is thoroughly tested.

10eMyrT:
Good morning! Considering the comments made to issue #289, this PR introduces testing for the document handler. This is a work in-progress, as I believe your comments will contribute to improving the code. Note that all the tests pass when I run `go test ./...` on my local machine. 
Closes #286


internal/handlers/telegram/handler_test.go
```diff
@@ -0,0 +1,40 @@
...
+
+	// Assign the MimeType(Optional) to FileID(Required) if it does not exists
+	if updateObj.Message.Document.MimeType == ""{
+		updateObj.Message.Document.MimeType = updateObj.Message.Document.FileID
+	}
+
+	// Change to FileName(Optional) is the Caption(Optional) is not given
+	if updateObj.Message.Caption == ""{
+		updateObj.Message.Document.FileName = "Random Text"
+	}
```
10eMyrT, line 33: I am cautious about Caption and FileName. They are both optional, meaning that a document may not have either a caption or a filename attached. The test might fail in that case. These are the two options we used in the `handler.go` file. Should we add another option or remove both Caption and FileName from how we format the string sent to IRC?

internal/handlers/telegram/handler_test.go
```diff
@@ -0,0 +1,40 @@
...
+
+func TestDocument(t *testing.T) {
+	correct := "Someone shared a file (Random File) on Telegram with caption: 'Random Caption'."
+	updateObj := &tgbotapi.Update{
+		Message: &tgbotapi.Message{
+			From: &tgbotapi.User{
+				FirstName: "Someone",
+			},
+			Document: &tgbotapi.Document{
+				FileID: "Random File",
```
Tjzabel, line 17: This file id will actually be the download link. So something like `"https://teleirc.com/file.txt"` as a fake download link would work.

internal/handlers/telegram/handler_test.go
```diff
@@ -0,0 +1,40 @@
...
+	"github.com/stretchr/testify/assert"
+	"testing"
+)
+
+func TestDocument(t *testing.T) {
+	correct := "Someone shared a file (Random File) on Telegram with caption: 'Random Caption'."
+	updateObj := &tgbotapi.Update{
+		Message: &tgbotapi.Message{
+			From: &tgbotapi.User{
+				FirstName: "Someone",
```
Tjzabel, line 14: So this works perfectly fine, but I would actually like to try to standardize on using `"test"` when we can as the `FirstName` or `UserName` of our test users. That way we don't need to think super hard about which test cases use which username.

This isn't documented anywhere, but I would like to strive for all of our tests to be laid out in the same fashion.

internal/handlers/telegram/handler_test.go
```diff
@@ -0,0 +1,40 @@
...
+			},
+			Caption: "Random Caption",
+		},
+	}
+	// Call the String() function which checks if a username(Optional) exists
+	updateObj.Message.From.UserName = updateObj.Message.From.String()
+
+	// Assign the MimeType(Optional) to FileID(Required) if it does not exists
+	if updateObj.Message.Document.MimeType == ""{
+		updateObj.Message.Document.MimeType = updateObj.Message.Document.FileID
```
Tjzabel, line 27: We don't want to set the mime type to anything if it doesn't exist. We just skip over it.

internal/handlers/telegram/handler_test.go
Tjzabel: I feel the caption is very useful to users on IRC. Even if there was no caption set, I feel that is still useful information to IRC users.

internal/handlers/telegram/handler_test.go
```diff
@@ -0,0 +1,40 @@
...
+				FirstName: "Someone",
+			},
+			Document: &tgbotapi.Document{
+				FileID: "Random File",
+			},
+			Caption: "Random Caption",
+		},
+	}
+	// Call the String() function which checks if a username(Optional) exists
+	updateObj.Message.From.UserName = updateObj.Message.From.String()
```
Tjzabel, line 23: Instead of explicitly setting the username to the first name if one doesn't exist, we should just call `user.String()` when we are building the final string to send to IRC.

internal/handlers/telegram/handler_test.go
10eMyrT: You are right. Standardizing would make it easier in the future.

internal/handlers/telegram/handler_test.go
10eMyrT: I moved it to the `handler.go`.

internal/handlers/telegram/handler.go
```diff
@@ -64,8 +64,12 @@ documentHandler receives a document object from Telegram, and sends
 a notification to IRC.
 */
 func documentHandler(tg *Client, u *tgbotapi.Message) {
-
-	formatted := u.From.UserName + " shared a file (" + u.Document.MimeType + ")"
+	formatted := u.From.String() + " shared a file ("
```
Tjzabel, line 67: Nice :+1: 
This is exactly how we want to use the `user.String()` method :100:

internal/handlers/telegram/handler.go
```diff
@@ -64,8 +64,12 @@ documentHandler receives a document object from Telegram, and sends
...
+	if u.Document.MimeType != "" {
+		formatted += u.Document.MimeType + ")"
+	} else {
+		formatted += u.Document.FileID + ")"
```
Tjzabel, line 71: So this is actually the whole issue with #159 
We explicity do _not_ want the unfiltered link to the document download, as this [displays the bot api token](https://github.com/RITlug/teleirc/issues/241). For now, I would recommend only showing the `MimeType` if it is available, and ignore the `FileID` entirely. 

This is something I forgot about in the first review.

internal/handlers/telegram/handler_test.go
```diff
@@ -0,0 +1,28 @@
+package telegram
+
+import (
+	tgbotapi "github.com/go-telegram-bot-api/telegram-bot-api"
+	"github.com/stretchr/testify/assert"
+	"testing"
+)
+
+func TestDocument(t *testing.T) {
+	correct := "test shared a file (https://teleirc.com/file.txt) on Telegram with caption: 'Random Caption'."
```
Tjzabel, line 10: Good, but due to my comment above, we don't actually want to show the file link. 
I would recommend just doing:
```
correct := "test shared a file on Telegram with caption: 'Random Caption'"
```
when there is no mimetype available, as we don't want to show the bot api token.

internal/handlers/telegram/handler_test.go
```diff
@@ -0,0 +1,28 @@
...
+	updateObj := &tgbotapi.Update{
+		Message: &tgbotapi.Message{
+			From: &tgbotapi.User{
+				FirstName: "test",
+			},
+			Document: &tgbotapi.Document{
+				FileID: "https://teleirc.com/file.txt",
```
Tjzabel, line 17: File ID is a required field, so this is good :+1:

internal/handlers/telegram/handler.go
10eMyrT: I see your point. For security reason, we definitely should not display the api token.

internal/handlers/telegram/documentHandler_test.go
```diff
@@ -0,0 +1,166 @@
...
+	tgbotapi "github.com/go-telegram-bot-api/telegram-bot-api"
+	"github.com/stretchr/testify/assert"
+	"testing"
+)
+
+/*
+TestDocument_plain checks the behavior of the document handlers if only required
+fields are available.
+*/
+func TestDocument_plain(t *testing.T) {
```
Tjzabel, line 13: I would actually prefer if we try not to use too many underscores in these function names. Me and @Zedjones have only been using underscores to explicitly state when the function booleans are on or off. 
```
func TestDocumentPlain
func TestDocumentCaption
func TestDocumentMime
func TestDocumentUsername
func TestDocumentNoCaption
func TestDocumentFull
```

10eMyrT: Right, I will solve that problem. Another point, should we all keep our test inside a single file `handlers_test.go` or for each handler, we create a test file?

PR #299, opened 2020-04-17
IRC kick handler

10eMyrT:
This PR adds the `Kick Handler` to IRC. There were a lot of possible messages to send, by I have used `kickFmt = "* %s kicked %s from %s"` because I believe it is simple and clear. Your comments are welcomed.


internal/handlers/irc/handlers.go
```diff
@@ -92,15 +93,30 @@ func quitHandler(c Client) func(*girc.Client, girc.Event) {
 	}
 }
 
+/*
+kickHandler handles the event when a user is kicked from the IRC channel.
+*/
+func kickHandler(c Client) func(*girc.Client, girc.Event) {
+	return func(gc *girc.Client, e girc.Event) {
+		c.logger.LogDebug("kickHandler triggered")
+		if c.TelegramSettings != nil && c.TelegramSettings.ShowKickMessage {
```
Zedjones, line 102: You can get rid of this check here
```go
if c.TelegramSettings != nil
```
It was unnecessarily added to the other handlers in a previous PR, I'm going to remove it from the other handlers in an upcoming PR

10eMyrT: Thanks for the review. I removed the check for `c.TelegramSettings != nil`.

Zedjones:
Functionality looks good, do you think it might be good to include the reason for the kick in the message? Opinions? @jwflory @Tjzabel

jwflory:
> Functionality looks good, do you think it might be good to include the reason for the kick in the message? Opinions? @jwflory @Tjzabel

I would like to include the kick reason in the message. This is how v1.x.x behaves I think.

Tjzabel:
@10eMyrT since we merged in @Zedjones 's changes, this PR now has issues due to the `kickHandler` not being under the `ClientInterface`.

10eMyrT:
Just push a new commit to solve the problem.

10eMyrT:
It checks out now. We are good to go.

PR #300, opened 2020-04-17
Write unit tests for join/part/quit handlers

### LINKED ISSUE #270: Write unit tests for join/part/quit handlers
# Summary

Unit tests do not currently exist for joining, parting, and quitting an IRC channel.

# Background

After writing these handlers, it is unclear if we have handled all edge cases. We should have unit tests to thoroughly test this functionality.

Full set of unit tests to catch as many edge cases as we can.

No alternatives.

# Details

The owner of this issue needs to think of what the joining, parting, and quitting handlers should support, and not support. Then, build unit tests around the desired and undesired functionality to make sure error handling exists in the appropriate locations.

# Outcome

Unit testing encompasses the join, part, and quit handlers. Allows us to feel safe knowing these handlers are thoroughly tested.

Zedjones:
This PR adds an interface for the IRC client and will include testing for the join, part, and quit handlers for IRC.

Resolves #270


Zedjones:
@jwflory Is there a way to disable Code Climate on a specific file or a file pattern? For example, two files that it's complaining about are auto-generated. Both are named `*_mock.go`.

10eMyrT:
You can solve the code climate issue by running `gofmt <file name>` for each of the file and commit. That is how I was able to deal with it in my PRs.

Tjzabel:
@jwflory CC seems to really not like duplicate code. Due to this being a unit test PR, there will be many "duplicate" code sections. How can we whitelist this PR for those issues?

internal/handlers/irc/handlers_test.go
```diff
@@ -0,0 +1,209 @@
+package irc
+
+import (
+	"testing"
+
```
Tjzabel, line 5: Is there a reason why there is a space between each import?

Zedjones:
> You can solve the code climate issue by running `gofmt <file name>` for each of the file and commit. That is how I was able to deal with it in my PRs.

This unfortunately won't work since none of the error are formatting related (and I also have `gofmt` set to run on a file whenever I save it).

Zedjones:
> @jwflory CC seems to really not like duplicate code. Due to this being a unit test PR, there will be many "duplicate" code sections. How can we whitelist this PR for those issues?

Ironically enough it's not complaining about the test files themselves (perhaps there is some sort of whitelist for that). It's complaining primarily about the auto-generated mocks from mockgen. Auto-generated files shouldn't really be considered as part of our code base, so optimally it would be nice to just remove them from Code Climate all together.

Tjzabel:
@Zedjones lol, I didn't see your comment earlier asking Justin the same thing :sweat_smile:

internal/handlers/irc/handlers_test.go
Zedjones: No there is not, lol. I meant to fix that, will do that soon.

Zedjones:
> From looking at this PR, everything _seems_ to check out. My only concern is how vastly different this method of testing is compared to how we are doing things over on the Telegram side. However, if this is the easiest method of testing IRC, then I'm all for it.
> 
> My only question is why there was an import statement with spaces between each import, but otherwise everything seems to check out. Thanks +1

Yeah, unfortunately since the handlers on the IRC side are side effect based and not purely input/output, the only way to effectively test them is to intercept those side effects (through mocking) and assure that they are what we expect. I could extract out the logic of every handler and only test on input/output but to me that would be more messy, especially since we need to do this style of testing to test the logic in `irc.go` and `telegram.go` anyway.

I think this will serve as a good example.

jwflory:
> @jwflory Is there a way to disable Code Climate on a specific file or a file pattern? For example, two files that it's complaining about are auto-generated. Both are named `mock_*.go`.

Fixed in #301.

jwflory:
Got a confirmation to merge this in IRC. Here it goes! :ocean:

PR #301, opened 2020-04-17
ci: Use exclude_patterns to avoid bad test data

jwflory:
@Zedjones had CodeClimate pick apart one of his pull requests because of
the interfaces used. This should get CodeClimate to calm down on the
repeated code by ignoring a few different patterns.

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -1,4 +1,5 @@
 ---
+version: "2"
 plugins:
   gofmt:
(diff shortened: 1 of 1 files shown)


jwflory:
Confirmed the fix. I rebased this on the changes in #300 and all of the flagged warnings were cleared:

https://codeclimate.com/github/RITlug/teleirc/pull/301

PR #304, opened 2020-04-18
docs: Officially license content CC BY-SA 4.0

jwflory:
This commit brings our documentation content under an explicit license,
instead of de facto being committed on the basis of the MIT License. A
Creative Commons Attribution-ShareAlike license is a better format for
written content and material. It is also a Free Culture license. I
figured this was the most open-ended license to use for this content.

I'm opening this PR as a proposal for discussion. This doesn't officially fit into our sprints but it is an important thing to be explicit about in our docs. I think this license serves better.

### DIFF
--- a/docs/LICENSE.txt
+++ b/docs/LICENSE.txt
@@ -1,0 +1,428 @@
+Attribution-ShareAlike 4.0 International
+
+=======================================================================
+
+Creative Commons Corporation ("Creative Commons") is not a law firm and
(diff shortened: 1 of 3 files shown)


jwflory:
I don't think this is a controversial change like #305, so I'm going to merge this through with one approval. If someone disagrees with the license choice, I'm happy to reopen this as a new issue.

PR #305, opened 2020-04-18
legal: Switch TeleIRC v2.0.0 license to GPL-3.0

jwflory:
This commit changes the license of RITlug/teleirc from the MIT License
to a GNU General Public License. Merging this commit to `master` is
conditional on the consent of all active team developers. This commit
message will explain the rationale for this change and why I think it is
important.

Share improvements back upstream
================================

The README always said to please share improvements back upstream, but
we are using the MIT License, perhaps the most permissive of all
software licenses. In respect to the volunteer developer time and
energy, I think the GPL is a better position of the TeleIRC project's
values as an open source project.

Copyleft > permissive
=====================

Most notably, we switch from a permissive license to a copyleft license.
If a permissive license says "this code is open source", a copyleft
license says "this code is open source and will stay open source". I
think the GPL is an effective license for TeleIRC because of the
fundamental copyleft essence of keeping changes to derivative works
under the same license


xforever1313:
I usually prefer the MIT or BSL license so then I can use libraries I contribute to or create at my job, like I do for some of the repos on my account (...yay corporate policies!).

However... teleirc is more application than it is library.  I don't foresee me ever linking with it.  So I'm fine with it.

jwflory:
_Discussed in [2020-04-25 developer meeting](https://bluejeans.com/s/WoGko)._

---

We reviewed the LGPL in the context of what TeleIRC is today and how the project is used from the last few years. The GPL makes more sense for TeleIRC as a standalone application. Like @xforever1313 said, there's not really a way to link to TeleIRC as a library.

So, now this PR changes the original MIT License with the GPLv3. Not much changes in the practicality of this change, as far as copyleft licenses go.

jwflory:
Thanks folks. All approving reviews on this pull request. I'm merging the pull request. From this point forward, we are a GPL project :)

PR #311, opened 2020-04-22
codeclimate: Add exclusion patterns for more files

jwflory:
This commit hides more files from CodeClimate that we don't want it to
measure. I know these files show up in the code progress reports, but it
is not clear to me if they are also included as untested source code for
our overall test coverage.


jwflory:
Thanks @10eMyrT! :raised_hands: Merging. :clapper:

PR #312, opened 2020-04-25
Write unit tests for message handler

### LINKED ISSUE #269: Write unit tests for message handler
# Summary

Unit tests do not currently exist for the message handler. 

# Background

After writing the message handler, it is unclear if we have handled all edge cases. We should have unit tests to thoroughly test the message handler functionality.

Full set of unit tests for the message handler to catch as many edge cases as we can.

No alternatives.

# Details

The owner of this issue needs to think of what the message handler should support, and not support. Then, build unit tests around the desired and undesired functionality to make sure error handling exists in the appropriate locations.

# Outcome

Unit testing encompasses the message handler. Allows us to feel safe knowing the message handler has been thoroughly tested.

Zedjones:
Added unit tests for the message handler and moved a string concatenation to the inside of a check since it was unused if the check failed.

Resolves #269

### DIFF
--- a/internal/handlers/irc/handlers_test.go
+++ b/internal/handlers/irc/handlers_test.go
@@ -314,3 +314,120 @@ func TestKickHandlerNoReason(t *testing.T) {
 		},
 	})
 }
+
+func TestMessageHandlerInBlacklist(t *testing.T) {
+	ctrl := gomock.NewController(t)
+
+	defer ctrl.Finish()
+
+	ircSettings := internal.IRCSettings{
+		IRCBlacklist: []string{
+			"SomeUser",
+		},
+	}
+
+	mockClient := NewMockClientInterface(ctrl)
+	mockLogger := internal.NewMockDebugLogger(ctrl)
+	mockClient.
+		EXPECT().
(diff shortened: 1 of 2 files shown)


internal/handlers/irc/handlers_test.go
```diff
@@ -314,3 +314,120 @@ func TestKickHandlerNoReason(t *testing.T) {
...
+}
+
+func TestMessageHandlerFull(t *testing.T) {
+	ctrl := gomock.NewController(t)
+
+	defer ctrl.Finish()
+
+	ircSettings := internal.IRCSettings{
+		IRCBlacklist: []string{},
+		Prefix:       "<<",
```
Tjzabel, line 399: Prefix & Suffix are just `"<"`, not the double `<<|>>`.

Zedjones: Resolved during meeting

jwflory:
In the developer meeting today, we decided to go ahead with this test as-is as another test case. Merge conflict fixed and tests passing. Merging. :clapper:

PR #317, opened 2020-04-26
systemd: Update service file for Go best practices

jwflory:
This PR includes two commits:

---

6a11c0a3ad325827527d68a0b1dbf2fe455283de — :outbox_tray: **Consolidate build/ + deployment/ dirs to misc/**

It was a little confusing to me to have multiple directories like this
for different things. I think it is better to unify all the non-Go build
assets and files into a single folder.

---

ebf3fb9752bd7cce968b5550f50a3487ccdabcf5 — :new: **systemd: Update service file for Go best practices**

This is based on changes introduced in jwflory/ansible-role-teleirc#1.

---


jwflory:
> Is this systemd service file a temporary measure of deployment, or are you planning on continuing use of systemd as a method of deployment after the v2.0 release?

I will use it for our documented Go server deployment. I will use this for my Ansible Role too.

Long-term, I was thinking of packaging it into Fedora / EPEL for wider distribution, but – one step at a time. :stuck_out_tongue_closed_eyes: 

Merging! :clapper:

PR #319, opened 2020-05-08
WIP: Testing Telegram Update Handler

10eMyrT:
This PR tests the update handler of Telegram. Just the functionalities of the update handler are tested, and what is done by the functions called in this handler is not addressed because each of those handlers have already been tested separately.


internal/handlers/telegram/handler_test.go
```diff
@@ -10,6 +10,160 @@ import (
...
+			return "documentHandler triggered"
+		case u.Message.Photo != nil:
+			return "photoHandler triggered"
+		default:
+			return "Triggered, but message type is currently unsupported"
+		}
+	}
+	return ""
+}
+
```
10eMyrT, line 36: I had to create this function here based on the handler itself, and remove the calls to the other methods to make it work.

internal/handlers/telegram/handler_test.go
```diff
@@ -10,6 +10,160 @@ import (
...
+						FirstName: "test",
+						UserName:  "testUser",
+					},
+				},
+			},
+		}
+	updates <- update
+
+	correct := "joinHandler triggered"
+	assert.Equal(t, correct, MockUpdateHandler(updates))
```
10eMyrT, line 54: This strategy works, and running `go test ./...` does complete successfully.

internal/handlers/telegram/handler_test.go
```diff
@@ -10,6 +10,160 @@ import (
...
+	// 		Suffix:          ">",
+	// 		ShowJoinMessage: true,
+	// 	},
+	// 	logger: internal.Debug {
+	// 		DebugLevel: true,
+	// 	},
+	// 	sendToIrc: func(s string) {
+	// 	},
+	// }
+	// updateHandler(clientObj, updates)
```
10eMyrT, line 68: Calling the update handler takes long time, and does not even complete when I run `go test ./...`. Your input on the reason is welcome, but I believe it stays in the loop waiting for another update.

Tjzabel: Most likely due to waiting for update events.

internal/handlers/telegram/handler_test.go
robbyoconnor: This looks incredibly fragile and prone to code rot if people forget this happened...

jwflory: @robbyoconnor Could you please clarify what you mean by if people forget this happens? This is in the context of unit tests, right?

robbyoconnor: I withdraw my comment.

internal/handlers/telegram/handler_test.go
Tjzabel: @Zedjones do you know why the client would be stalling on the channel here?

jwflory:
@Tjzabel @Zedjones @10eMyrT Is this PR in scope for the release today? Trying to decide if I bump to the `v2.x.x` milestone or keep it on target for today.

Zedjones:
I think a better way to go about this might be to use `var` at the top level of the module to mock out each handler individually. This current method does not allow for protections against regressions when adding functionality to the `updateHandler` and it also doesn't account for the logging being done in the actual `updateHandler`.

[Here's](https://github.com/Zedjones/RedirectProtect/blob/3d20fc6a012a04e7f4ffab810163a562e9b6e878/internal/time_checking.go) an example of how to use `var` at the top level. And [here's](https://github.com/Zedjones/RedirectProtect/blob/3d20fc6a012a04e7f4ffab810163a562e9b6e878/internal/time_checking_start_time_check_test.go) an example of how to use that in a test.

jwflory:
_Discussed in [2020-05-23 meeting](https://bluejeans.com/s/kF2he)._

---

## Release-blocking: _No_

We discussed in the meeting, and @Zedjones summarized the meeting feedback in his comments above. We decided we won't try to rush or expedite these tests for the release if they are not fully polished yet.

Will follow up after v2.0.0 release.

10eMyrT:
Thanks for the reviews. I will get back to it this week and push an update as soon as finished.

Tjzabel:
Closing due to inactivity.

PR #320, opened 2020-05-09
Added initial Dockerfile

Zedjones:
Added a Dockerfile using Alpine Linux and some build scripts around it.
I chose Alpine because, unlike Node, Go is A) a compiled language and B) a statically compiled language.
As such, our binary currently only relies on these shared libraries:
```
~/Documents/RITLug/teleirc/deployments/container dockerfile
❯ ldd teleirc
        linux-vdso.so.1 (0x00007ffc2c126000)
        libpthread.so.0 => /usr/lib/libpthread.so.0 (0x00007f1abb5d0000)
        libc.so.6 => /usr/lib/libc.so.6 (0x00007f1abb40a000)
        /lib64/ld-linux-x86-64.so.2 => /usr/lib64/ld-linux-x86-64.so.2 (0x00007f1abbe07000
```
And, as you can see, I did have to do a symlink in the container because the linux shared object was expected to be named differently. 
However, after building, this image is only 15.7 MB, which is incredibly nice.

To build the image, you can run the `deployments/container/build_image.sh` script from anywhere and it'll create an image tagged `teleirc`. Alternatively, to run the Docker Compose example, you need to `cd` into `deployments/container` and run `docker-compose -f docker-compose.yml.example build teleirc`.


deployments/container/Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM alpine
+
+RUN mkdir /lib64 && ln -s /lib/libc.musl-x86_64.so.1 /lib64/ld-linux-x86-64.so.2
+
+COPY teleirc /opt/teleirc/
+
+ENTRYPOINT [ "/opt/teleirc/teleirc" ]
```
robbyoconnor, line 7: Er -- just use a [multistage build](https://docs.docker.com/develop/develop-images/multistage-build/)

Also, add an unprivileged user.

deployments/container/Dockerfile
```diff
@@ -0,0 +1,15 @@
+FROM golang:1.13 AS teleirc-builder
+WORKDIR /opt/teleirc
```
robbyoconnor, line 2: LABEL maintainer="TeleIRC Team <someemail@somedomain.com>"

PR #322, opened 2020-05-17
Add ZWSP between Telegram usernames

Tjzabel:
# What This Is
If a user has accounts on both IRC and Telegram with the same username, it is easy to ping oneself across platforms each time a message is sent. A zero-width space (ZWSP) is used to add an invisible differentiation between the usernames on these platforms to prevent cross-platform username pinging.

## Developer Details
`ResolveUserName` is now used instead of `user.String()` for retrieving username information.
For additional details, see: https://github.com/42wim/matterbridge/issues/175.

### DIFF
--- a/internal/config.go
+++ b/internal/config.go
@@ -28,6 +28,7 @@ type IRCSettings struct {
 	Suffix              string   `env:"IRC_SUFFIX" envDefault:">"`
 	ShowJoinMessage     bool     `env:"IRC_SHOW_JOIN_MESSAGE" envDefault:"true"`
 	ShowLeaveMessage    bool     `env:"IRC_SHOW_LEAVE_MESSAGE" envDefault:"true"`
+	ShowZWSP            bool     `env:"IRC_SHOW_ZWSP" envDefault:"true"`
 	NickServPassword    string   `env:"IRC_NICKSERV_PASS" envDefault:""`
 	NickServService     string   `env:"IRC_NICKSERV_SERVICE" envDefault:""`
 	EditedPrefix        string   `env:"IRC_EDITED_PREFIX" envDefault:"[EDIT] "`
--- a/internal/handlers/telegram/helpers.go
+++ b/internal/handlers/telegram/helpers.go
@@ -13,8 +13,7 @@ func GetUsername(u *tgbotapi.User) string {
 	}
 	// Add ZWSP to prevent pinging across platforms
 	// See https://github.com/42wim/matterbridge/issues/175
-	username := u.UserName[:1] + "" + u.UserName[1:]
-	return username
+	return u.UserName[:1] + "" + u.UserName[1:]
 }
 
 /*
@@ -37,6 +36,5 @@ func ZwspUsername(u *tgbotapi.User) string {
 	}
 	// Add ZWSP to prevent pinging across platforms
 	// See https://github.com/42wim/matterbridge/issues/175
-	username := u.UserName[:1] + "" + u.UserName[1:]
-	return username
+	return u.UserName[:1] + "" + u.UserName[1:]
 }
--- a/internal/handlers/telegram/helpers.go
+++ b/internal/handlers/telegram/helpers.go
@@ -5,9 +5,9 @@ import (
 )
 
 /*
-ResolveUserName does basic cleanup if a user does not have a username on Telegram.
+GetUsername returns a Telegram user's username if one is set, or first name otherwise.
 */
-func ResolveUserName(u *tgbotapi.User) string {
+func GetUsername(u *tgbotapi.User) string {
 	if u.UserName == "" {
 		return u.FirstName
 	}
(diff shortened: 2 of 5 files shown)


internal/handlers/telegram/helpers.go
```diff
@@ -5,12 +5,25 @@ import (
...
 on Telegram, we first need to check to see if they have one set.
 */
-func GetUsername(u *tgbotapi.User) string {
+func GetFullUsername(u *tgbotapi.User) string {
 	if u.UserName == "" {
 		return u.FirstName
 	}
-	return u.FirstName + " (@" + u.UserName + ")"
+	return u.FirstName + " (@" + u.UserName[:1] + "" + u.UserName[1:] + ")"
 }
```
jwflory, line 29: I am a little confused, what makes `ResolveUserName` and `GetFullUsername` different here?

internal/handlers/telegram/helpers.go
```diff
@@ -5,12 +5,25 @@ import (
 )
 
 /*
-GetUsername returns the name and username of a user. Since usernames are optional
+ResolveUserName does basic cleanup if a user does not have a username on Telegram.
+*/
+func ResolveUserName(u *tgbotapi.User) string {
```
jwflory, line 10: I guess it goes without saying that the ZWSP will only ever appear in IRC clients and not in Telegram clients. That is a correct assumption, right?

internal/handlers/telegram/helpers.go
Tjzabel: `ResolveUserName` is the simple version of `GetFullUsername`. It is used any time we need a person's username. 

`GetFullUsername` is *only* used in the `join/part` handlers, which gets both the person's first name, *and* their Telegram username. This is the same setup we used in the NodeJS version.

internal/handlers/telegram/helpers.go
Tjzabel: Yes, this only appears on IRC, and does not affect Telegram.

internal/handlers/telegram/helpers.go
jwflory: @Tjzabel Makes sense, thanks for the answer. Should have read your code comments too, I didn't see them in the diff. I have two new pieces of feedback, will open a new thread for these.

internal/handlers/telegram/helpers.go
jwflory, line 10: Since this method inserts a ZWSP in the username, I think the function name and comment should be more clear that the character is inserted.

Also an aside, but this is probably a good place to insert a boolean check later, where if TeleIRC is configured to have ZWSP off, it will not be inserted?

internal/handlers/telegram/helpers.go
```diff
@@ -5,12 +5,25 @@ import (
...
+GetFullUsername returns the name and username of a user. Since usernames are optional
 on Telegram, we first need to check to see if they have one set.
 */
-func GetUsername(u *tgbotapi.User) string {
+func GetFullUsername(u *tgbotapi.User) string {
 	if u.UserName == "" {
 		return u.FirstName
 	}
-	return u.FirstName + " (@" + u.UserName + ")"
+	return u.FirstName + " (@" + u.UserName[:1] + "" + u.UserName[1:] + ")"
```
jwflory, line 28: Dumb question but I want to be explicit. The `(@username)` part is only returned if a user has a username set, otherwise it is just their first name. Right?

internal/handlers/telegram/helpers.go
Tjzabel: @jwflory oh, you're right! I completely forgot to add the boolean with the ZWSP check. I will get to that tomorrow.

internal/handlers/telegram/helpers.go
Tjzabel: yep, since usernames are not required by Telegram, this will only return the First name (which is required) if no username is set.

internal/handlers/telegram/helpers.go
jwflory: @Tjzabel Cool. This is the last thing on my feedback list before this is good to merge. :+1:

internal/handlers/telegram/helpers.go
```diff
@@ -5,12 +5,38 @@ import (
...
+username.
+*/
+func ZwspUsername(u *tgbotapi.User) string {
+	if u.UserName == "" {
+		return u.FirstName
+	}
+	// Add ZWSP to prevent pinging across platforms
+	// See https://github.com/42wim/matterbridge/issues/175
+	username := u.UserName[:1] + "" + u.UserName[1:]
+	return username
```
Zedjones, line 41: Kinda silly, but there's really no reason to have an extra variable here. You can just change this to 
```go
return u.UserName[:1] + "" + u.UserName[1:]
````

Comment (Tjzabel) [reply internal/handlers/telegram/helpers.go]: fixed.

**Comment by jwflory** [comment]
Looks good to me. I'm going to merge this in. Thanks for getting this done @Tjzabel @Zedjones!

:clapper:

File: internal/config.go [hunk 5]
```diff
@@ -28,6 +28,7 @@ type IRCSettings struct {
 	Suffix              string   `env:"IRC_SUFFIX" envDefault:">"`
 	ShowJoinMessage     bool     `env:"IRC_SHOW_JOIN_MESSAGE" envDefault:"true"`
 	ShowLeaveMessage    bool     `env:"IRC_SHOW_LEAVE_MESSAGE" envDefault:"true"`
+	ShowZWSP            bool     `env:"IRC_SHOW_ZWSP" envDefault:"true"`
```
jwflory, line 31: Note: This needs to go into the `env.example` and our config glossary too.

PR #325, opened 2020-05-23
docs: Add specific deployment steps (closes #297)

### LINKED ISSUE #297: Documentation improvement re install/deploy
this is a copypasta of pertinent parts of a conversation between myself and @Tjzabel and @jwflory .

i didn't know that "Read the Docs" had a version selector. in all of the years that i've used RTD, i'd not noticed it until @Tjzabel provided someone else a link to a specific-version document yesterday. realizing that was incredibly helpful. 

but also, the "Quick Install" and "How to deploy TeleIRC" documents seem... inadequate on their own. IME/IMO most people do not separate the concepts of installation from deployment. like at all. those of you who work with node/go every day may well keep those as separate concepts, but i certainly don't. if you continue to have them as separate documents, please at least describe their interaction on each page; i've worked in ops for nearly 20 years, so i probably understand their relationship better than others might. 

also, the systemd service file linked from the docs is missing. i _eventually_ found it, just now, by cloning the repo and looking at commit history to see if it had changed or where it had gone. turns out it's just been moved, as part of a commit that references Go (and hence v2) (it's now at deployments/systemd/teleirc.service), despite its content still referencing v1.

## Maintainer checklist (_edited_)

- [x] ~~Add callout for older docs versions~~ 1bf5ad6c054ec768a4a25eb12167ff58a477a238
- [x] ~~Merge "Quick install" / "How to deploy" doc into a single Quick Start page~~ ae6adeeeb5979560a39391796e9c17e9923a2a8f
- [ ] Correctly document systemd / bare metal approach to deployment
- [x] ~~Make sure all version strings are updated and correct~~ 71177fab7713aed63136f37325b7cf427733ab92

jwflory:
These steps are a quick runthrough that basically explains how the
Ansible Role works. But this gives a quick and easy pathway for people
who are running TeleIRC to deploy it on a bare-metal Linux distribution
with `systemd`. We can try to support other use cases later, but I think
this is sufficient level of depth for release day.

Closes #297.

### DIFF
--- a/docs/user/quick-start.md
+++ b/docs/user/quick-start.md
@@ -152,7 +152,7 @@ This section is only required if you are building a binary from source:
 1. Clone repository (`git clone https://github.com/RITlug/teleirc.git`)
 1. Enter repository (`cd teleirc/`)
 1. Install dependencies (`go install`)
-1. Build binary (`go build ./...`)
+1. Build binary (`go build cmd/teleirc.go`)
 
 #### Configuration
 
--- a/docs/user/quick-start.md
+++ b/docs/user/quick-start.md
@@ -18,10 +18,11 @@ Note this does not apply to v1.x.x releases; see the [v1.3.4 documentation][1].
     1. [Configure Imgur Image Upload (IIU)](#configure-imgur-image-upload-iiu)
 1. [Deployment Guide](#deployment-guide)
     1. [Run binary](#run-binary)
-        1. [Requirements](#requirements)
-        1. [Install dependencies](#install-dependencies)
+        1. [Pre-requirements](#pre-requirements)
+        1. [Build TeleIRC](#build-teleirc)
         1. [Configuration](#configuration)
         1. [Start bot](#start-bot)
+            1. [Example Linux setup](#example-linux-setup)
     1. [Run container](#run-container)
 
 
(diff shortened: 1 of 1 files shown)


docs/user/quick-start.md
```diff
@@ -130,15 +133,26 @@ There are two ways to deploy TeleIRC persistently:
...
+
+#### Build TeleIRC
 
-1. Clone the repository (`git clone https://github.com/RITlug/teleirc.git`)
+This section is only required if you are building a binary from source:
+
+1. Clone repository (`git clone https://github.com/RITlug/teleirc.git`)
+1. Enter repository (`cd teleirc/`)
 1. Install dependencies (`go install`)
+1. Build binary (`go build ./...`)
```
Tjzabel, line 155: I feel this should be `go build cmd/teleirc.go` as _this_ is what will ultimately create an executable `teleirc` binary in the current working directory.

docs/user/quick-start.md
```diff
@@ -151,15 +165,67 @@ See [_Config file glossary_][11] for detailed information.
...
+To start the bot, you need to consider the following factors:
+
+1. Where will the binary go?
+1. Where is your config file on the system?
+1. How will you automate the bot to start-up automatically after a system reboot?
+
+##### Example Linux setup
+
+**NOTE**:
+_Looking for an easier way?_
```
Tjzabel, line 182: Is this really an easier way? Would it be worthwhile to simply state that `go run cmd/teleirc.go`, or if built, `./teleirc` will work just as well?

docs/user/quick-start.md
jwflory: @Tjzabel Fixed in commit c78d50314c83e5e51e6e77fd764576a54bd8b976, thanks for catching that.

docs/user/quick-start.md
jwflory: @Tjzabel This is the setup guide I would want for myself. It is probably more simple to use `go run cmd/teleirc.go` but I would rather encourage best practices. Plus I was starting to think about packaging this in Fedora/EPEL at some point in the future, so eventually I could deprecate this in support of something like package manager steps.

I'm not convinced the way I wrote it is best, but I'd like to try it out for now and see if folks find it useful or helpful.

jwflory:
Thanks @Tjzabel! Merging this in. v2.0.0 is officially done! :tada:

PR #334, opened 2020-07-24
Fix Telegram zero-width space

### LINKED ISSUE #328: Zero width space (ZWSP) implementation is faulty
# Summary

The ZWSP is supposed to add an unseen space between letters in a username to prevent users across platform from pinging themselves. The implementation TeleIRC currently has is faulty; it does not achieve these desired results. 

## How to reproduce?

1. Have the same username on both IRC & Telegram
2. Send a message on one platform to another
3. See yourself pinged

## Expected behavior

Users who have the same username on IRC & Telegram should not inadvertently be pinging themselves.

Tjzabel:
## Overview

This issue finally puts an end to the ZWSP issues. Users who have the same Telegram and IRC nicknames should now be able to safely message from each platform, and avoid pinging oneself. 

Closes #328 

### Details
Telegram usernames now contain a ZWSP when sent over to IRC to prevent self-pinging:
```
<t<200b>estUser> Hello
```
This resolves into:
```
<testUser> Hello
```

## Questions
1. Do we want this option to have a toggle?


jwflory:
@Tjzabel Since it is a ZWSP, I might not be able to clearly see the change happening or how to validate the fix:

Screenshot of helpers.go git diff from GitHub Pull Request interface

Could you provide some context into what your change is doing to fix the bug?

> **@Tjzabel**:
> Do we want this option to have a toggle?

Based on feedback we got in the community chat, I think yes. I believe we should default the ZWSP to present though. I hypothesize there are fewer use cases for this to be disabled (but enough that it is important to make it configurable).

Tjzabel:
@jwflory I agree it should at least be on by default. I can work on creating a toggle for this option then.

I understand it's hard to view via the diff itself, which is why I included how it looks in the original comment. Basically, if you copy/paste the `""` you see in the `diff` into a terminal, you will be able to see a `<200b>` zwsp character

jwflory:
Updated this Pull Request to a draft PR.

Tjzabel:
@jwflory may I ask why this was moved to a draft PR?

jwflory:
Oh sorry! Is it ready for final review? I forgot where the status was on this.

The CI failing is an issue I need to look at again, right? :thinking:

Tjzabel:
This PR is effectively complete, as it passes all tests locally for me, and runs as it should. The main issue is the CI itself is borked due to possibly needing a newer version. I will open up a new issue for that

internal/handlers/telegram/handler.go
```diff
@@ -126,7 +146,12 @@ documentHandler receives a document object from Telegram, and sends
 a notification to IRC.
 */
 func documentHandler(tg *Client, u *tgbotapi.Message) {
-	username := GetUsername(u.From)
+	username := ""
+	if tg.IRCSettings.ShowZWSP {
+		username = ZwspUsername(u.From)
+	} else {
+		username = GetUsername(u.From)
+	}
```
jwflory, line 154: This code feels repetitive since it is repeated many times over in the other handlers. This adds additional testing overhead from my point-of-view. Is it possible to add a helper function somewhere else that will return the correct username based on whether the Zwsp setting is enabled?

internal/handlers/telegram/helpers.go
```diff
@@ -23,7 +21,20 @@ func GetFullUsername(u *tgbotapi.User) string {
...
+func GetFullUserZwsp(u *tgbotapi.User) string {
+	if u.UserName == "" {
+		return u.FirstName
+	}
+	// Add ZWSP to prevent pinging across platforms
+	// See https://github.com/42wim/matterbridge/issues/175
+	return u.FirstName + " (@" + u.UserName[:1] + "​" + u.UserName[1:] + ")"
 }
 
 /*
```
jwflory, line 40: Maybe adding the check function somewhere in this file?

jwflory (changes requested):
@Tjzabel This looks great, and I am super excited for this to merged and deployed :heart_eyes: 

I noticed some of the code seemed repetitive and left more detailed feedback in my review. I focused on the implementation and did not review the tests very carefully. It would be nice if @Zedjones, @kennedy, or @10eMyrT could check out the tests.

This is almost ready from my POV! Thanks so much for working on this.

internal/handlers/telegram/handler.go
Tjzabel: I actually thought for a while about just moving this into a separate helper function in the `helpers.go` file. My _only_ reasoning for doing it this way was it requires adding several more layers of complexity into the helpers file, but it definitely makes the `handler.go` file much cleaner, so I will look into splitting that out more.

jwflory: @Tjzabel Got it, makes sense. I would be OK addressing this in a future PR if it is complex to do right now. I would rather get this merged and deployed, and take time for polish and small improvements once we hit a regular cadence of meetings again.

jwflory (approved):
Changing my review to an approval based on @Tjzabel's follow-up comment. We can split out the function to `helpers.go` when we have more capacity.

jwflory:
@Tjzabel I'll leave it to you to merge if you think it is ready. But it looks good on my side. :slightly_smiling_face:

Tjzabel:
@jwflory I was planning to work on the function separation but my bandwidth turned out to be quite low :sweat_smile: 
I will do a final review of this PR, and submit an `enhancement` issue for creating that zwsp-check function.

PR #336, opened 2020-10-02
ci: Upgrade golangci-lint to v1.31.0

jwflory:
This is to fix it breaking our C.I. pipeline because apparently, this
version doesn't exist anymore? This is where caching of build assets
would be nice actually.

Related to #334.


Tjzabel (approved):
Looks good to me! Thanks for doing this :+1:

PR #339, opened 2020-10-10
No need for space in defaults, already in handler

### LINKED ISSUE #333: Telegram messages prepend an extra space
# Summary

When sending a message from TG -> IRC, and extra space is prepended in TG messages.

# Standard debugging steps

## How to reproduce?

1. Run TeleIRC in `HEAD`
2. Send a message on Telegram
3. See there are two spaces at the beginning of the message text instead of 1

## Expected behavior

Only a single space is seen at the start of TG -> IRC messages.

**Screenshots**:

image

thenaterhood:
It's been a while, so may as well throw my hat in the ring again in the spirit of Hacktoberfest :)

Fixes #333. The handler already appends a space after the suffix, so there isn't a need for a space in the default suffix.

_Alternatively_ we could change the formatting in the handler and leave the space here, not sure if we have a preference.

### DIFF
--- a/internal/config.go
+++ b/internal/config.go
@@ -41,7 +41,7 @@ type TelegramSettings struct {
 	Token               string `env:"TELEIRC_TOKEN,required"`
 	ChatID              int64  `env:"TELEGRAM_CHAT_ID,required"`
 	Prefix              string `env:"TELEGRAM_MESSAGE_PREFIX" envDefault:"<"`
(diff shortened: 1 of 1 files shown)


jwflory (approved):
Hey @thenaterhood, nice to see you here! :grinning: This change is pretty simple so I am going to merge it through. Thanks for fixing this!

PR #340, opened 2020-10-10
Don't forward messages with a configurable prefix

### LINKED ISSUE #201: Feature request: filter messages starting with a prefix
An IRC channel I use for work has a bot set up to log all messages except the ones starting with a specific prefix (`[off]`), for privacy concerns.

Currently, that is stopping us from using TeleIRC, as we don't want those messages logged elsewhere.

A feature to filter outgoing messages that start with a prefix would be greatly appreciated (hopefully not just for our use-case) and fairly easy to implement.

I'm happy to open a PR if you are interested in this feature.

thenaterhood:
Fixes #201

### DIFF
--- a/docs/user/config-file-glossary.rst
+++ b/docs/user/config-file-glossary.rst
@@ -69,6 +69,9 @@ IRC settings
 ``IRC_SHOW_ZWSP=true``
     Prevents users with the same Telegram and IRC username from pinging themselves across platforms.
 
+``IRC_NO_FORWARD_PREFIX="[off]"``
+    A string users can prefix their message with to prevent it from being relayed across the bridge. Removing this option or setting it to "" disables it.
+
 
 *****************
 Telegram settings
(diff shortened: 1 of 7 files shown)


thenaterhood:
I added this for both sides, since I've seen concern from folks who occasionally don't want their messages leaving Telegram.

Tjzabel (changes requested):
@thenaterhood awesome! Since you added to the config file, can you please add this info into the `docs/` as well?

Tjzabel (approved):
LGTM

jwflory:
I tested this locally and it works like a charm! I followed @MonkeyDo's use case with the `[off]` prefix:

Screenshot of a conversation over TeleIRC - from the IRC side

Screenshot of a conversation over TeleIRC - from the Telegram side

Going to merge this. :clapper: Thanks @thenaterhood! :raised_hands:

Issue #228, opened 2020-01-01
Deploy docker images to dockerhub


cintiadr:
# Summary
Firstly, thanks so much for the tool. 

You already provide the [dockerfiles](https://github.com/RITlug/teleirc/tree/master/images), I was wondering if you'd consider also deploying the images to dockerhub as part of your release. 

# Background

**Is your improvement related to a problem? Please describe**:
Right now, in order to use your docker image, I have to clone your repo, build the image and push to dockerhub under my own name. 

**Describe the solution you'd like**:
Ideally images in dockerhub, that are updated on each release. 

**Describe alternatives you've considered**:
Every person deploying their own version. 

# Outcome

No need to clone or build docker images to use teleirc.

jwflory:
@cintiadr Thanks for opening this issue. While the Dockerfiles do work, they will be rewritten as part of the v2.0 release detailed in #163. Once we release v2.0, this is something I am happy to do unless @Tjzabel or @Zedjones have a strong preference for it.

For now, I am keeping this on our v2.0 backlog. Thanks! :+1:

jwflory:
_Discussed in [2020-04-11 meeting](https://bluejeans.com/s/RViMY)._

---

Marked as blocked on #193. Will revisit in a May 2020 sprint.

robbyoconnor:
Oh Hi @cintiadr !

oliveratgithub:
I tested various `teleirc` Docker images with a `docker-compose.yml`, here's a summary and some obvervations:

| Docker image | Release | Support / Compatibility |
|---|---|---|
| [burke/teleirc](https://hub.docker.com/r/burke/teleirc/tags) | -<br>(9 years ago) | 👍 **Works out-of-the-box**, `linux/amd64` only, logging is very basic, and  some newer Settings are unsupported, e.g.:<br>❌ `TELEGRAM_MESSAGE_PREFIX`<br>❌ `IRC_NO_FORWARD_PREFIX`<br>❌ `SHOW_JOIN_MESSAGE`<br>❔ `IRC_BOT_IDENT`<br>✅ `IRC_MESSAGE_PREFIX` |
[...]

PR #344, opened 2020-10-11
Added a DisconnectHandler that sends a message to Telegram if IRC Dies

### LINKED ISSUE #96: Send Telegram notification when connection to IRC is lost/restored
# Summary

Send notification from Telegram bot to group when IRC connection is lost and when it is restored

# Background

Sometimes the bot may crash, restart, or otherwise not be connected. In IRC, this is clear because the bot has to join the channel and have proper permissions to speak. In Telegram, there is no indication when the connection is broken. It might be helpful to prevent a [situation like this](https://www.youtube.com/watch?v=k0pL_CsK3Dk)…

# Details

I'm not sure what our current options are to get the state of the IRC connection. If it's easy to get, then we should have an event handler for when the connection disconnects and reconnects after the bot's initial start-up.

# Outcome

Less confusion when a connection drops and cross-bridge conversation is interrupted (and messages will never be relayed automatically later)

xforever1313:
Resolves #96

When the IRC Connection's DISCONNECTED event is triggered, the bot will now send a "lost connection" message to the Telegram Channel.  Specifically, the message will be "Lost connection to '<channel>' on '<server>'".  It will only show this message if the "SHOW_DISCONNECT_MESSAGE" setting is set to true.

This was tested by adding a firewall rule on my test PC to block the IRC port, and sure enough, a few seconds later, we get the error message in Telegram.

...
Side-Note: It appears as though if the IRC connection does die, the process also dies.  We get this in the Console:

```
ERROR: 2020/10/11 17:22:02 read tcp 192.168.15.174:56524->195.148.124.79:6697: wsarecv: An established connection was aborted by the software in your host machine.
ERROR: 2020/10/11 17:22:02 read tcp 192.168.15.174:56524->195.148.124.79:6697: wsarecv: An established connection was aborted by the software in your host machine.
```

There is no reconnection logic built into Teleirc other than the thing launching (e.g. systemd) it auto-restarting it, so that's probably fine (if not, may be worth a new issue).


Tjzabel (approved):
Checks out for me!

robbyoconnor (approved):
Looks **_GREAT_** to me!

robbyoconnor:
Shall I merge?

robbyoconnor:
I merged.

jwflory:
This came up in the dev chat, but thought I would put it here for posterity too. If the IRC connection dies, it is OK if the process also dies. This is handled through the systemd unit file we ship upstream to automatically restart the process when it dies. Also, later on, once we revisit containers ( #228 ), k8s or OpenShift can spin up new pods if one connection dies.

Tjzabel:
@xforever1313 actually, can you please add the `IRC_DISCONNECT` option to the config glossary in the docs as well?

jwflory:
Oops, I always forget the docs for changes to the config options :see_no_evil:

PR #345, opened 2020-10-15
Update version field

Tjzabel:
Now that we are on `v.2.0.1` we need to make sure our versioning reflects that. 

After this quick PR, I am going to create an issue to see if there is a better way we can track the current TeleIRC version.


robbyoconnor (approved):
LGTM

robbyoconnor:
Triggering CI build.

PR #348, opened 2020-10-17
Added Chaskis IRC Bot Framework as a user of TeleIRC

xforever1313:
Chaskis does use TeleIRC to bridge Telegram and IRC.

(Admittedly, its probably the _only_ user in the IRC channel at times, but it is there! 🤣 )


Tjzabel (approved):
:100:

jwflory:
Awesome, thanks for adding it here @xforever1313! :tada:

PR #352, opened 2020-10-25
Resolves #346 (Determine how best to display TeleIRC version flag)

### LINKED ISSUE #346: Determine how best to display TeleIRC version flag
# Summary

Every time we make a new release of TeleIRC, we need to make sure we are also updating the TeleIRC versioning flag.

# Background

Users were confused on our latest `v.2.0.1` release, as our versioning flag still says `v2.0`.

We need to make sure our versioning flag correctly reflects the most up to date release. Therefore, this should always be the latest change we make before we cut a new release.

None. This needs to happen.

Zedjones:
Test using the `./build_binary.sh` and then verify that the version displayed by `./teleirc --version` matches the latest version tag and current commit (v2.0.1-9-ge003bdd)


cmd/teleirc.go
```diff
@@ -13,14 +13,11 @@ import (
...
 
-const (
-	version = "v2.0.1"
-)
-
 var (
 	flagPath    = flag.String("conf", ".env", "config file")
 	flagDebug   = flag.Bool("debug", false, "disable debugging")
 	flagVersion = flag.Bool("version", false, "displays current version of TeleIRC")
+	version     string
```
jwflory, line 20: Out of curiosity, why declare the `version` as a `string` here? Where is this referenced, or where does it get called?

jwflory (approved):
Simple enough to me. It is nifty that this is not that difficult to pull off in Go with the build-time variables. I'm glad this will make future debugging down to the exact commit much easier!

Thanks for sleuthing this one @Zedjones. Merging. 🎬

cmd/teleirc.go
Zedjones: It's injected at compile time by the shell script, that's why it's such a hack haha.

PR #361, opened 2020-12-17
Add Imgur API support for images sent from Telegram

jwflory:
This is a work-in-progress Pull Request to add support for Imgur to TeleIRC. Currently this is a bit buggy and will segfault when you upload an image to Telegram, but it does compile successfully. There is more context in the commit messages if you want to see how I've approached this.

Going to leave this here for feedback if anyone sees something I am missing since I am a Go noob. Otherwise I am aiming to gradually work on this through December.


Tjzabel:
I've done some updates on this PR. As of right now, the following features are working:

1. TeleIRC grabs the correct Telegram photo link (shows bot API token)
2. Prints the URL out the standard out
3. Has the general structure of what is needed

Tjzabel:
## PR Update
This PR is now feature-complete. When Telegram users send images, they will get converted to an imgur link, and sent over to IRC. 

### Changes
* Imgur uploading is fully operational and no longer optional
* Default Go version is now 1.14
* **PhotoHandler test cases are skipped for the time being while we figure out the best approach to testing this functionality**
* All messages now make use of an actual `\u200b` ZWSP instead of a hidden one. 

Thanks all for being patient for the implementation of this feature :smile:

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/ritlug/teleirc
 
-go 1.12
+go 1.14
```
jwflory, line 3: Yay!

internal/handlers/telegram/handler_test.go
```diff
@@ -696,7 +699,7 @@ func TestMessageRandomWithNoForward(t *testing.T) {
 			ChatID: 100,
 		},
 		IRCSettings: &internal.IRCSettings{
-			ShowZWSP: false,
+			ShowZWSP:        false,
```
jwflory, line 702: Tab character here?

internal/config.go
```diff
@@ -60,8 +60,7 @@ type TelegramSettings struct {
 
 // ImgurSettings includes settings related to Imgur uploading for Telegram photos
 type ImgurSettings struct {
-	UseImgurForImageLinks bool   `env:"USE_IMGUR_FOR_IMAGE" envDefault:"true"`
-	ImgurClientID         string `env:"IMGUR_CLIENT_ID" envDefault:"7d6b00b87043f58"`
+	ImgurClientID string `env:"IMGUR_CLIENT_ID" envDefault:"7d6b00b87043f58"`
```
jwflory, line 63: Is this a new API token? I assume we can distribute this as a TeleIRC-default while encouraging others to rotate with their own API token.

internal/handlers/telegram/handler.go
```diff
@@ -120,9 +120,13 @@ photoHandler handles the Message.Photo Telegram object. Only acknowledges Photo
 exists, and sends notification to IRC
 */
 func photoHandler(tg *Client, u tgbotapi.Update) {
+	link := uploadImage(tg, u)
 	username := GetUsername(tg.IRCSettings.ShowZWSP, u.Message.From)
-	formatted := username + " shared a photo on Telegram with caption: '" +
-		u.Message.Caption + "'"
+	caption := u.Message.Caption
+	if caption == "" {
+		caption = "Untitled Message"
```
jwflory, line 127:
```suggestion
		caption = "No caption provided."
```

jwflory: I feel like this is more direct English for folks on the IRC side

env.example
```diff
@@ -76,5 +76,4 @@ SHOW_DISCONNECT_MESSAGE=true
 #                                                                             #
 ###############################################################################
 
-USE_IMGUR_FOR_IMAGES=true
```
jwflory, line 79: I assume this is leftover from v1?

jwflory (commented):
Overall, this looks good to me and I am happy to see it merge. I left some review comments and offered one suggestion on a logging message, so nothing major. This makes sense to me.

However, since this is "technically" my Pull Request, GitHub is not allowing me to approve this Pull Request. So, @Tjzabel, you will have to approve your own changes. 😂

internal/config.go
Tjzabel: This line just gets rid of the `UseImgurForImageLinks` line. We already use a default `ImgurClientID`.

internal/handlers/telegram/handler.go
Tjzabel: makes sense

internal/handlers/telegram/handler_test.go
Tjzabel: `gofmt` reformatting.

env.example
Tjzabel: Yes, we never required imgur, but at this point making imgur optional just makes for more unnecessary overhead.

Tjzabel (approved):

PR #365, opened 2021-03-05
fix: Fix broken link to "Config file glossary" in env.example file.

da-edra:
1. Why is this change neccesary?
The current link to the "Config file glossary" documentation is broken.

Screenshot to showcase the issue:
Screenshot that shows a message saying "Sorry, this page does not exist yet".

2. How does it address the issue?
By adding the right URL in the env.example file.

3. What side effects does this change have?
None.


Tjzabel (approved):
Woops! Thanks for the catch!

PR #367, opened 2021-03-07
only forward messages we care about

### LINKED ISSUE #366: ban evasion bug. user can pm the bot and send messages to irc without being in the group
# Summary

A user who isn't part of the group can send messages to the IRC side by privately messaging the bot on telegram

# Standard debugging steps

## How to reproduce?

message the bot in private on telegram

## Expected behavior

Bot should ignore the private message

**Screenshots**:

## Stacktrace

# Other details

I haven't tested if the same behaviuor occurs on the IRC side as well

thenaterhood:
Should fix #366.

Looks like we just were blindly forwarding messages to IRC without checking where they came from. The opposite direction (IRC to Telegram) appears not to have this problem.

### DIFF
--- a/internal/handlers/telegram/handler.go
+++ b/internal/handlers/telegram/handler.go
@@ -61,6 +61,12 @@ func messageHandler(tg *Client, u tgbotapi.Update) {
 		return
 	}
 
+	// Don't forward messages to IRC that didn't come from the
+	// chat we're bridging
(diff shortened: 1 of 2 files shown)


xforever1313 (approved):

Tjzabel:
Thanks for taking care of this!

Issue #51, opened 2018-04-30
Include limited characters from Telegram replies in relayed IRC message


bexelbie:
The teleboto bot in #fedora-zh does to interesting things.  Notice this exchange from the IRC side:

<teleboto> [alick9188] @bexelbie You're living in China?
<teleboto> [bexelbie] 「Re alick9188: @bexelbi...」Nope. I’ll be in Beijing for LC3 though.

First, the name from telegram is in square brackets, making it more obvious it is passed through.

Second, it quotes part of the replied to message to give greater context.

WDYT?

jflory7:
I see this issue as a UI/UX enhancement for two things:

1. Place Telegram usernames between square brackets (which I forgot, we actually can do already – there are [config values](https://github.com/RITlug/teleirc/blob/devel/.env.example#L13) for username prefix/suffix)
2. Quote a number of characters for messages from a replied message on Telegram and relay it over it IRC

I will update this accordingly.

## Implementation ideas

Quoting messages should be a configurable options. If set to true, there should also be a configuration value for the number of characters for quoting messages. Assume a default of maybe 15 characters.

When configured accordingly, the messages appear like this:

```
<alick9188> @bexelbie You're living in China?
<bexelbie> 「Re alick9188: @bexelbi...」Nope. I’ll be in Beijing for LC3 though.
```

@robbyoconnor @repkam09 @thenaterhood @xforever1313 Do any of you have thoughts or opinions on this?

xforever1313:
...I swear we had a way in teleirc to determine if a message from telegram was a reply or not...  Did I dream that 0_o?  That would be step one.

Step two would be getting the message that was replied to.  Its been a while since I looked at the Telegram Bot API... but I would assume it would include either the replied message as a string, a message object, or an ID on a message object.  If its a string or message object, that's not too bad to grab, but an ID could get tricky as we'll have to make a second request to get the string of the message of that ID (or heaven forbid do caching).

The rest is straightforward, pre-pending the IRC message with the characters from the config, and making a substring of the message that was replied to.

[...]

PR #371, opened 2021-04-20
Add the username when a TG user reply to someone

mscherer:
This pull request try to convey the fact that someone replied to someone else when discussing, by adapting the message to the IRC convention.

If user A reply to user B on TG, this will appear like:
```
tg-bot>  <A> B: A very valid point, but I still think chocolate and soy sauce are incompatible
```

Instead of:

```
tg-bot>  <A> A very valid point, but I still think chocolate and soy sauce are incompatible
```

This is quite useful since people sometime answer a few hours later and it is hard to follow from irc.

I didn't test the patch (yet), but it compile.

### DIFF
--- a/internal/handlers/telegram/handler.go
+++ b/internal/handlers/telegram/handler.go
@@ -56,6 +56,7 @@ Telegram update into a simple string for IRC.
 */
 func messageHandler(tg *Client, u tgbotapi.Update) {
 	username := GetUsername(tg.IRCSettings.ShowZWSP, u.Message.From)
+	replyUsername := ""
 
 	if tg.IRCSettings.NoForwardPrefix != "" && strings.HasPrefix(u.Message.Text, tg.IRCSettings.NoForwardPrefix) {
 		return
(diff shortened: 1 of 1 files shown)


mscherer:
So the patch is not working great:
```
23:06:13| @tg-bot> <Misc> %!(EXTRA string=(not a great success))
```

I guess there is some issue with the TG API that I overlooked

jwflory:
Hey @mscherer, thanks for opening this! 🙌🏻 

Not sure if you already saw this, but there is documentation for the [`tgbotapi` package](https://pkg.go.dev/github.com/go-telegram-bot-api/telegram-bot-api). It might help clarify what is going on here.

I am also pinging folks who have worked on these files before, in case they have a better idea of what is going on. @kennedy @Tjzabel @Zedjones

thenaterhood:
Looks like it works fine to me!
```
<t​henaterhood>t​henaterhood: foobar
```
I'd recommend adding a space between the username and the reply-to name but otherwise this looks great.

Tjzabel:
Hello! As @jwflory points out, this issue is directly related to #51. This PR is a great start, but I am not sure this is the method that we want to use to handle Telegram replies. I feel adding context to the replies (so IRC users know exactly which TG message was replied to) could be very beneficial.

Do you have any interest in continuing this work with #51? If not, I am going to close this PR and make a note of this in the other issue so we may use it as inspiration for when it gets picked up in the future.


Thanks :smile:

mscherer:
It make sense, I agree a more structured way to deal with answer could be needed. However, I do not think I understand the context of #51 enough to provides anything, I just went for a quick fix, I do not have the bandwidth for a bigger refactor for the time being.

Tjzabel:
> It make sense, I agree a more structured way to deal with answer could be needed. However, I do not think I understand the context of #51 enough to provides anything, I just went for a quick fix, I do not have the bandwidth for a bigger refactor for the time being.

That is fully understandable. I love the idea of this PR and I feel it will be extremely helpful in completing #51. I'm going to close this issue and use pieces of this PR as a baseline for that issue :+1:

PR #372, opened 2021-05-13
add "!" for people without a username

randomhydrosol:
this is an approach we are using because anyone can put <firstname> as somebody else's <username> and effectively impersonate them

### DIFF
--- a/internal/handlers/telegram/helpers.go
+++ b/internal/handlers/telegram/helpers.go
@@ -9,7 +9,7 @@ GetUsername takes showZWSP condition and user then returns username with or with
 */
 func GetUsername(showZWSP bool, u *tgbotapi.User) string {
 	if u.UserName == "" {
-		return u.FirstName
+		return "!" + u.FirstName
 	}
 	if showZWSP {
 		return ZwspUsername(u)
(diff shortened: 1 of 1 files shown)


Tjzabel:
Thanks for this PR @anupritaisno1!. At this moment I am unsure if this issue is something that would be beneficial to the main TeleIRC branch.

RZS110 (approved):
I think it's a good idea

Tjzabel:
I love the basic idea of this issue. However, I am not sure an IRC user is going to fundamentally understand that a Telegram user with a name of `!User` is one without a username set. Due to this, I believe this PR may bring more confusion than it fixes.

jwflory:
Hi @randomhydrosol, I tried to rebase this Pull Request on latest changes in `main`, but I do not think you allowed maintainers to make edits to your Pull Request. Could you please rebase this PR on latest changes in `main` since our CI pipeline is working again?

If all the tests pass, we can reopen this for discussion in the Telegram group on a path forward here. Otherwise we can close this out as `stale`.

Tjzabel:
Hello,

I think this change will cause more confusion than it solves in the long-run. We cannot expect IRC users to immediately understand what a `!` means in front of a Telegram name, and therefore I do not think this change should be merged into the main codebase. 

I think this is a great idea overall, and you are welcome to make use of it in your own TeleIRC fork 🙂

PR #389, opened 2021-12-19
Default SHOW_ACTION_MESSAGE to true

fgaz:
Like it's written in the docs https://docs.teleirc.com/en/latest/user/config-file-glossary/#telegram-settings
As discussed on telegram.


Tjzabel (approved):
Thanks for catching this! Approving 👍

PR #399, opened 2022-04-26
Update systemd deployment resources

half-duplex:
- Use systemd-tmpfiles and systemd-sysusers
- Use service template for ease of running multiple instances
- Simplify installation commands (`chown`/`chmod` -> `install`)

Based on my [changes](https://github.com/half-duplex/aur-teleirc/commit/92fa701fd53b5df03148bdc8e69ff84c7aa0e1a6) to the [AUR package](https://aur.archlinux.org/packages/teleirc) for teleirc.

I wrote the download links for 2.2.1, if the release ends up being 2.3.0 or something those should be updated.

### DIFF
--- a/deployments/systemd/teleirc.sysusers
+++ b/deployments/systemd/teleirc.sysusers
@@ -1,0 +1,1 @@
+u teleirc - "TeleIRC Service"
--- a/deployments/systemd/teleirc.tmpfiles
+++ b/deployments/systemd/teleirc.tmpfiles
@@ -1,0 +1,1 @@
+d /etc/teleirc 550 teleirc teleirc - -
--- a/deployments/systemd/teleirc@.service
+++ b/deployments/systemd/teleirc@.service
@@ -1,12 +1,12 @@
 [Unit]
-Description=TeleIRC: Telegram + IRC bridge bot
+Description=TeleIRC: Telegram + IRC bridge bot for %I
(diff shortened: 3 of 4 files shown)


thenaterhood:
In general this looks good to me and this is an elegant way to handle multiple instances. My only question is what are the backwards-compatibility implications for existing installs, if any? I'm a little rusty on systemd but I see the systemd service has a parameterized config path with these changes.

half-duplex:
Thanks!

People who used the current install instructions will have a service file that refers to `/etc/teleirc/env` and `/usr/local/bin/teleirc`. If they only replace the binary for updates, it'll continue working exactly the same.
If they want to continue using `/etc/teleirc/env` without renaming it but use the new service file etc, they should be able to just disable/delete the old one and `systemctl enable --now teleirc@env.service`.

I should also note that I haven't tested the selinux bits as I don't have an appropriate system handy.

thenaterhood (approved):

Tjzabel:
This _seems_ good to me, but I would personally like to defer to @jwflory on this if possible as he's the one who wrote the original systemd deployment.

Tjzabel (approved):

jwflory (commented):
A late review but this looks great! A much easier way of deploying and managing multiple bots. I learned something new. Thanks for this @half-duplex.

PR #401, opened 2022-05-29
Imgur: Log deletion links

half-duplex:
The Imgur upload process currently discards the deletion links. Since images are uploaded anonymously, deletion links are the only practical way to remove them. This PR logs them as INFO.

I don't really speak golang, not sure if `!= ""` is the right conditional for `omitempty`

### DIFF
--- a/internal/handlers/telegram/imgur.go
+++ b/internal/handlers/telegram/imgur.go
@@ -79,5 +79,10 @@ func getImgurLink(tg *Client, tgLink string) string {
 		tg.logger.LogError("Could not retrieve imgur json data:", err)
 	}
 
+	if resp.Data.Deletehash != "" {
+		deleteLink := "https://imgur.com/delete/" + resp.Data.Deletehash
+		tg.logger.LogInfo("Deletion link for", resp.Data.Link, "is", deleteLink)
+	}
+
 	return resp.Data.Link
 }


internal/handlers/telegram/imgur.go
```diff
@@ -79,5 +79,10 @@ func getImgurLink(tg *Client, tgLink string) string {
 		tg.logger.LogError("Could not retrieve imgur json data:", err)
 	}
 
+	if resp.Data.Deletehash != "" {
```
thenaterhood, line 82: You got it, this is correct. The struct member `Deletehash` is not a pointer, so if it's not in the response JSON then it will be initialized to an empty string.

thenaterhood:
This looks good to me.

I think in general this is something we should flesh out more at some point into a more official feature, but having the deletion links in some form at all is a huge improvement, even if it's just to log them.

thenaterhood (approved):

jwflory (approved):
The change seems simple to me and I concur with @thenaterhood's feedback. Does anyone else from @RITlug/teleirc-developers-commit-access want to review this before we merge? I haven't done much work with Go lately.

Tjzabel (approved):
Looks good to me! This is a really simple addition that is very nice for sysadmins who are asked to take down an image.

PR #404, opened 2022-10-02
Added Location handler

xforever1313:
Added the ability for if someone on Telegram shares their location, it will be forwarded to IRC.

This setting is defaulted to disabled for privacy reasons.

On the IRC side, this is what it will look like:

```
[16:11.27] <sethtestbot> xforever1313 shared their location: (43.0845274, -77.6781174)
```


xforever1313:
Hmm, looks like one of the CI jobs failed, but I don't think it was related to my changes?

```
Error: you must supply a CC_TEST_REPORTER_ID ENV variable or pass it via the -r flag
```

jwflory:
> Hmm, looks like one of the CI jobs failed, but I don't think it was related to my changes?

There is a secret key injected only on git branches from the RITlug remote. If you use a feature branch on the RITlug remote as a committer, this issue won't happen in your PRs. If it only fails at the reporting step, then it's a safe-to-ignore failed check.

docs/user/config-file-glossary.rst
```diff
@@ -107,6 +107,10 @@ Message settings
 ``IRC_SHOW_ZWSP=true``
     Prevents users with the same Telegram and IRC username from pinging themselves across platforms.
 
+``IRC_SHOW_LOCATION_MESSAGE=false``
+    If a user shares their location on Telegram, this will forward the GPS coordinates
+    of their location to IRC if set to true.
```
jwflory, line 112: Are we still following the [one-sentence-per-line convention](https://asciidoctor.org/docs/asciidoc-recommended-practices/#one-sentence-per-line)? I can't remember if that is documented. If so, 111-112 be changed to one line for an easier `git diff` if/when the docs are updated in the future.

```suggestion
If a user shares their location on Telegram, this will forward the GPS coordinates of their location to IRC if set to true.
```

internal/handlers/telegram/handler_test.go
```diff
@@ -882,3 +882,64 @@ func TestMessageFromWrongTelegramChat(t *testing.T) {
...
+
+func TestLocationHandlerWithLocationEnabled(t *testing.T) {
+	testUser := &tgbotapi.User{
+		ID:        1,
+		UserName:  "test",
+		FirstName: "testing",
+		LastName:  "123",
+	}
+
+	location := &tgbotapi.Location{
```
jwflory, line 894:
```suggestion

       // https://pkg.go.dev/github.com/go-telegram-bot-api/telegram-bot-api#Location
       location := &tgbotapi.Location{
```

internal/handlers/telegram/handler_test.go
```diff
@@ -882,3 +882,64 @@ func TestMessageFromWrongTelegramChat(t *testing.T) {
...
+		Latitude:  43.0845274,
+		Longitude: -77.6781174,
+	}
```
jwflory, line 897: 😂 https://www.google.com/maps/place/43.0845274,-77.6803061

jwflory (approved):
Looks good to me and tests are simple. 👍🏻

internal/handlers/telegram/handler_test.go
```diff
@@ -882,3 +882,64 @@ func TestMessageFromWrongTelegramChat(t *testing.T) {
...
+
+	correct := "test shared their location: (43.0845274, -77.6781174)."
```
jwflory, line 899: Is anyone up for Crossroads? 😉

internal/handlers/telegram/handler.go
```diff
@@ -184,3 +188,26 @@ func documentHandler(tg *Client, u *tgbotapi.Message) {
 
 	tg.sendToIrc(formatted)
 }
+
+/*
+locationHandler receivers a location object from Telegram, and sends
```
Tjzabel, line 193: "receivers" -- should be "receives"

This is the only issue, so still approving 👍

Tjzabel (approved):

PR #405, opened 2022-10-06
Build static binary

weezel:
This pull request compiles binary to static so it's easier to move binary around different machines having the same architecture. Also, removed debug symbols and build related paths from the binary that reduced the size from ~8MB down to 5MB.

### DIFF
--- a/build_binary.sh
+++ b/build_binary.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 CURRENT_VERSION=`git describe`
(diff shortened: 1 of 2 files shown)


jwflory (approved):
I tested two binaries, one before and one after the patch in a test channel, and everything worked well. I appreciate a leaner binary image! Thanks for the optimization, @weezel.

weezel:
> I tested two binaries, one before and one after the patch in a test channel, and everything worked well. I appreciate a leaner binary image! Thanks for the optimization, @weezel.

My pleasure :blush: If I missed something, please let me know.

Tjzabel (approved):

PR #406, opened 2022-11-03
Fix implicit memory aliasing in a loop.

weezel:
Otherwise the code will point into the same memory address during the loop.


okutoo:
image

PR #420, opened 2023-11-28
Fix long lines being cut by the IRC server

user890104:
Split long messages to runes, according to IRC server's max line length


Tjzabel:
Hello, can you please add a unit test for if a long message goes over the max length?

Tjzabel:
It actually looks like we may be able to get native support from https://github.com/lrstanley/girc/pull/43. Going to keep this PR open until I can look into this further.

PR #426, opened 2024-10-28
Reformat GitHub Actions

Tjzabel:
GitHub Actions revamped to move `go.yml` to `main.yml` as it is most accurately the main GHA we want to run. Previous `main.yml` is now `publish_docker_image.yml`. Currently, this only publishes our TeleIRC image to our private GHCR. 

Fixes Go builds to properly pull correct Go versions.


Tjzabel:
Goal is to eventually remove circleCI entirely. I believe we need to add code coverage to GHA then should be good.

PR #454, opened 2026-05-25
Pass through insecure TLS options to IRC connection

StevenLooman:
The insecure TLS options in the environment were never used/passed through to the IRC connection. This patch allows for insecure connections.


StevenLooman:
Not sure why the go_1-17 check fails, seems to be something with the CI itself. Other PRs also seem to be having this issue.

Also, Go 1-17 seems [end of life](https://endoflife.date/go) for a while now.

